PR #2, opened 2019-09-13
Drop signal validation test

thxCode:
[golang#33311](https://github.com/golang/go/pull/33311) has been proved and merged into golang repo, we don't need to validate again.


loganhz:
LGTM, removed a test which cannot run with DinD

PR #6, opened 2020-07-30
removed sorting of args

sachinmsft:
When we pass the args to app from the wins client they are getting sorted before passing to the app on server.
that can cause problem if app is not using named args.

e.g. I am passing below args from the wins client
8080 TCP 80 10

but they got converted as below after sorting.
10 80 8080 TCP

here we are receiving args in different order than the order we have passed resulting in breaking the app args parsing.

I could not figure out any reason of sorting the args. please let me know if i have missed any point here.

### DIFF
--- a/cmd/cmds/flags/list_value.go
+++ b/cmd/cmds/flags/list_value.go
@@ -1,7 +1,6 @@
 package flags
 
 import (
-	"sort"
 	"strings"
 
 	"github.com/urfave/cli"
@@ -28,7 +27,6 @@ func (f *listValue) Get() []string {
 		return nil
 	}
 	ret := strings.Split(f.String(), valueSeparator)
-	sort.Strings(ret)
 	return ret
 }
 


thxCode:
@sachinmsft I'm sorry to reply so late, do you execute something like `wins cli prc --args c b a`? I think you are right, we don't need to sort the input arguments, at least, we should not do this in `args` argument.

thxCode:
The code is LGTM, but the CI is not happy, @sachinmsft  can you trigger it again? cc @luthermonson

sachinmsft:
@thxCode the problem seems to be related to installing cygwin on docker image.

luthermonson:
I'll work on this @sachinmsft and this will get in on the next release

aiyengar2:
@sachinmsft I'm going to drop this PR in favor of https://github.com/rancher/wins/pull/17/commits/8f6f69fbda77a2e0564783f79ebb75dbb75e7577, which upgrades wins and fixes the Cygwin install issue noted in https://github.com/rancher/wins/pull/6#issuecomment-710265275 by updating the Dockerfile.

Feel free to re-open this PR or comment on https://github.com/rancher/wins/pull/17 if there are any additional concerns!

aiyengar2:
@sachinmsft fyi, these changes have been merged. Will be released in a v0.0.5 of wins

Issue #13, opened 2020-12-23
Add support for publishing ports


aiyengar2:
Wins currently supports exposing ports on the host via an [`--exposes "{TCP|UDP}:{Port}"`](https://github.com/rancher/wins/blob/7c2d5528151cb63355615e1ee02bd59380c1c1e2/cmd/client/process/run.go#L35-L39) option on a running process.

There should be a similar option available for `--publishes "{TCP|UDP}:{Port}"` that allows a container to publish a port available on the host network as a container port. This would allow privileged containers using wins to avoid exposing a port on the host itself when it only needs to be exposed on a container-level (e.g. inter-Pod communication for Kubernetes vs. inter-Node communication).

sowmyav27:
Will wait for this https://github.com/rancher/rancher/issues/28327 epic to be fixed to test this issue out.

thehejik:
I see the support is already incorporated in the wins-0.1.0 binary. Anyway I'd like to test it. 
@aiyengar2 could you please provide me a simple test with description?

I found some bits related to `--exposes` but not for `--publishes` eg.:
https://github.com/rancher/wins/blob/master/tests/integration/docker/nginx.ps1
https://github.com/rancher/wins#readme

This is help from wins command on node wiht rancher-wins-upgrader deployed:

```
administrator@IP-AC1F0A6D C:\etc\rancher>wins cli proxy --help
NAME:
   rancher-wins cli proxy - Set up a proxy for a port via Rancher Wins (note: only TCP is supported)

USAGE:
   rancher-wins cli proxy [command options] [arguments...]

OPTIONS:
   --publish value  [required] [list-argument] Publish a port or a range of ports, e.g.: TCP:443 TCP:80-81 (note: only TCP is supported)
   --proxy value    [optional] Specifies the name of the proxy listening named pipe (default: "rancher_wins_proxy")


administrator@IP-AC1F0A6D C:\etc\rancher>wins --version
rancher-wins version v0.1.0
```

thehejik:
We realized it was actually tested within v2 monitoring testcases in Monitoring+v2+Windows+Support document. Closing.

PR #14, opened 2021-02-09
Add support for wins cli proxy

aiyengar2:
This PR adds support for `.\wins.exe cli proxy --publish TCP:<PORT>`. It is dependent on a prior PR to upgrade wins as a whole.

Only the last commit on this PR is relevant for this change.

Related Issue: https://github.com/rancher/wins/issues/13

Dependent on PRs:
- [x] https://github.com/rancher/wins/pull/17


luthermonson (approved):

cmd/client/proxy/publish.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to parse --publish %s: %v", publishPorts, err)
+	}
+	_proxyPorts = ports
+
+	return nil
+}
+
+func parsePublishes(publishPorts []string) (ports []int, err error) {
+	for _, pub := range publishPorts {
```
nickgerace, line 58: A bit confused about this for loop: why do we return an error when one `pub` does not satisfy what we are looking for? Should we continue the loop.

cmd/server/config/config.go
```diff
@@ -59,6 +63,11 @@ func (c *WhiteListConfig) Validate() error {
 			return errors.New("could not accept blank path as process white list")
 		}
 	}
+	for _, proxyPort := range c.ProxyPorts {
+		if proxyPort < 0 || proxyPort > 0xFFFF {
```
nickgerace, line 67: What does this check ensure?

nickgerace (approved):
LGTM barring a few comments

cmd/client/proxy/publish.go
aiyengar2: Because then the input for wins cli proxy is malformed, which means we should fail out or have unintended consequences.

e.g. imagine if I did
`wins cli proxy TCP:9090TCP:8080 UDP:9090`

You wouldn't want to continue with just `UDP:9090`

cmd/server/config/config.go
aiyengar2: a valid port is provided. Ports are limited to 16 bit integers, but the input takes in any size int

cmd/client/proxy/publish.go
nickgerace: Entirely agree. Thanks for explaning

cmd/server/config/config.go
nickgerace: Nice. Glad to see a thorough check lol

PR #17, opened 2021-02-10
Update Go version, remove vendor, and misc. upgrades

aiyengar2:
As part of the changes necessary for adding the `wins cli proxy` command, a ton of small changes to wins w.r.t. the Go version and dependencies and build workflow needed to be changed first.

I've split all of these non-proxy related changes from https://github.com/rancher/wins/pull/14 to this PR in order to make it easier to review.

Also takes care of escaped quote support for https://github.com/rancher/wins/issues/9.

Related Issue: https://github.com/rancher/wins/issues/13, https://github.com/rancher/wins/issues/9

PR #18, opened 2021-02-18
Support watching config for changes

aiyengar2:
This PR adds support for watching the path of the config file as well to automatically update wins when it is on watchingMode.

This is a necessary change to support updating the wins service whenever the config only gets modified (but not the binary). 

Only the last commit on this PR is relevant for this change.

Dependent on PRs:
- [ ] https://github.com/rancher/wins/pull/17


aiyengar2:
While this PR is still conceptually valid, it is being dropped as a dependency for https://github.com/rancher/wins/pull/14 in favor of a PR to add `wins upgrade`

aiyengar2:
Closing this PR in favor of wins upgrade, can be re-opened if necessary.

PR #19, opened 2021-02-23
Add wins up[grade]

aiyengar2:
This PR adds support for running `.\wins up[grade]`, which runs a Powershell script within wins that restarts the service with a different wins server binary.

This is an alternative to the current mechanism for upgrades that wins provides wherein it watches for a path and utilizes checksums to see whether the service should be modified / restarted and can also be used as an alternative method to enable wins on a Windows host (e.g. you can just run `.\wins up` instead of `.\wins srv app run --register; Start-Service -Name rancher-wins`).

The primary use case of this feature is to support a new Helm chart that will allow users to maintain wins configs / wins versions across entire clusters using just a Helm chart for configuration. This Helm chart uses wins (cli) to pass wins (upgrade) to wins (server) in order to update wins (server) on the host on demand. More succinctly it runs:
```powershell
# Transfer the generated config to the directory
Transfer-File -Src "c:\scripts\config" -Dst "c:\host\etc\rancher\wins\config"
# Force a restart of wins server, which will naturally use the newly generated config
wins.exe cli prc run --path=$pathToWinsBinary --args="up --wins-args=`'--register --config=$pathToWinsBinaryDir\config`'"
```

More information on this can be found in https://github.com/rancher/charts/pull/996.

Only the last commit of this PR is related to changes from this PR.

Related Issue: https://github.com/rancher/wins/issues/23

Dependent on PRs:
- [x] https://github.com/rancher/wins/pull/17

### DIFF
--- a/cmd/upgrade/internal/powershell/powershell.go
+++ b/cmd/upgrade/internal/powershell/powershell.go
@@ -1,0 +1,63 @@
+package powershell
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// Sourced from https://github.com/flannel-io/flannel/blob/d31b0dc85a5a15bda5e606acbbbb9f7089441a87/pkg/powershell/powershell.go
+
+//commandWrapper ensures that exceptions are written to stdout and the powershell process exit code is -1
+const commandWrapper = `$ErrorActionPreference="Stop";try { %s } catch { Write-Host $_; os.Exit(-1) }`
+
+// RunCommand executes a given powershell command.
+//
+// When the command throws a powershell exception, RunCommand will return the exception message as error.
+func RunCommand(command string) ([]byte, error) {
+	cmd := exec.Command("powershell.exe", "-NoLogo", "-NoProfile", "-NonInteractive", "-Command", fmt.Sprintf(commandWrapper, command))
+
+	stdout, err := cmd.Output()
+	if err != nil {
+		if cmd.ProcessState.ExitCode() != 0 {
+			message := strings.TrimSpace(string(stdout))
+			return []byte{}, errors.New(message)
+		}
+
+		return []byte{}, err
+	}
+
+	return stdout, nil
+}
+
+// RunCommandf executes a given powershell command. Command argument formats according to a format specifier (See fmt.Sprintf).
+//
+// When the command throws a powershell exception, RunCommandf will return the exception message as error.
+func RunCommandf(command string, a ...interface{}) ([]byte, error) {
+	return RunCommand(fmt.Sprintf(command, a...))
+}
+
+// RunCommandWithJSONResult executes a given powershell command.
+// The command will be wrapped with ConvertTo-Json.
+//
+// You can Wrap your command with @(<cmd>) to ensure that the returned json is an array
+//
+// When the command throws a powershell exception, RunCommandf will return the exception message as error.
+func RunCommandWithJSONResult(command string, v interface{}) error {
+	wrappedCommand := fmt.Sprintf(commandWrapper, "ConvertTo-Json (%s)")
+	wrappedCommand = fmt.Sprintf(wrappedCommand, command)
+
+	stdout, err := RunCommandf(wrappedCommand)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(stdout, v)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/cmd/upgrade/internal/powershell/powershell.go
+++ b/cmd/upgrade/internal/powershell/powershell.go
@@ -54,8 +54,7 @@ func RunCommandWithJSONResult(command string, v interface{}) error {
 		return err
 	}
 
-	err = json.Unmarshal(stdout, v)
-	if err != nil {
+	if err = json.Unmarshal(stdout, v); err != nil {
 		return err
 	}
 
--- a/cmd/upgrade/upgrade.go
+++ b/cmd/upgrade/upgrade.go
@@ -1,0 +1,130 @@
+package upgrade
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,130 @@
...
+var _upgradeFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:  "binary",
+		Usage: "[optional] Name of binary in working directory to upgrade to",
+		Value: os.Args[0], // default behavior is to use current binary
+	},
+	cli.StringFlag{
+		Name:  "wins-args",
+		Usage: "[optional] Arguments to pass onto wins srv app run",
+		Value: "--register",
```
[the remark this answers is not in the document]
aiyengar2: done

cmd/upgrade/internal/powershell/powershell.go
```diff
@@ -0,0 +1,63 @@
+package powershell
```
luthermonson, line 1: couldnt this replace /pkg/powershells?

cmd/upgrade/internal/powershell/powershell.go
```diff
@@ -0,0 +1,63 @@
...
+	wrappedCommand := fmt.Sprintf(commandWrapper, "ConvertTo-Json (%s)")
+	wrappedCommand = fmt.Sprintf(wrappedCommand, command)
+
+	stdout, err := RunCommandf(wrappedCommand)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(stdout, v)
+	if err != nil {
```
nickgerace, line 58: nit: `if err = json.Unmarshal(stdout, v); err != nil {`

cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,112 @@
...
+		if ($currBinPath -ne $newBinPath) {
+			Copy-Item -Recurse -Force -Path $newBinPath -Destination $currBinPath | Out-Null
+		}
+	} else {
+		$currBinPath = $newBinPath
+	}
+
+	Invoke-Expression "& $currBinPath srv app run --register $winsArgs"
+	Start-Service -Name rancher-wins;
+	Write-Host "Upgraded rancher-wins"
```
nickgerace, line 53: Is this whitespace going to be an issue?
EDIT: and for other places in the file

cmd/upgrade/upgrade.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	binPath := os.Args[0]
+	winsArgs := cliCtx.String("wins-args")
+	out, err := powershell.RunCommandf(upgradePS1Fmt, binPath, winsArgs)
+	if len(out) > 0 {
+		logrus.Debugf("logs from upgrade.ps1 \n%s\nEOF", out)
+	}
+	if err != nil {
+		logrus.Errorf("upgrade failed, attempting to ensure rancher-wins service is not stopped: %v", err)
+		out, restartServiceErr := powershell.RunCommand(restartServicePS1)
```
nickgerace, line 101: What if the restart is successful? Should the upgradeAction return the error?

nickgerace (approved):
LGTM, but the `_upgradeAction` comment is one I'm curious about

cmd/upgrade/internal/powershell/powershell.go
aiyengar2: I don't think it can without some serious refactoring. `pkg/powershell` executes Powershell commands via a goroutine (so it supports executing a Powershell command in parallel) vs. these are blocking Powershell calls

cmd/upgrade/internal/powershell/powershell.go
aiyengar2: Since this is sourced from https://github.com/flannel-io/flannel/blob/d31b0dc85a5a15bda5e606acbbbb9f7089441a87/pkg/powershell/powershell.go, I'm just going to leave this comment as is instead of modifying it for the nit.

cmd/upgrade/upgrade.go
aiyengar2: When I tried it out in the rancher-wins-upgrader chart with a local copy, it did not cause any issues. So I think the whitespace should be fine

cmd/upgrade/upgrade.go
aiyengar2: Yup, the error from upgradeAction is logrus.Errorfed in the line you highlighted and returned. The only error that is never returned and only logged is restartServiceErr.

There is a bug in the logic here though; the second if should be checking if restartServiceErr is not nil and the log statement is wrong. Will make the update to the PR in another commit.

cmd/upgrade/internal/powershell/powershell.go
nickgerace: Don't want to edit the upstream code being brought in?

cmd/upgrade/upgrade.go
nickgerace: Agreed, just wanted to double check

cmd/upgrade/upgrade.go
nickgerace: > the second if should be checking if restartServiceErr is not nil and the log statement is wrong
Yep! this is what I was referring to. Apologies for lack of clarity :)

cmd/upgrade/internal/powershell/powershell.go
aiyengar2: If it's something that impacts the functionality of the code, yes. But for a nit change, I don't think we need to modify it if it's copied from upstream to make it easier to rebase in the future.

cmd/upgrade/upgrade.go
aiyengar2: updated, resolving this

cmd/upgrade/internal/powershell/powershell.go
aiyengar2: eh actually, I just added the fix. Might as well get it in 😅

luthermonson (approved):

Issue #24, opened 2021-02-25
Create build images for rancher/wins on Dockerhub


aiyengar2:
Requires moving over the logic from https://github.com/rancher/wmi_exporter-package to create docker files that just package wins.exe and enabling Drone builds for rancher/wins images on Dockerhub.

Will be used by Helm chart created in https://github.com/rancher/wins/issues/20.

PR #25, opened 2021-03-09
Publish docker images for wins

aiyengar2:
Related Issue: https://github.com/rancher/wins/issues/24


scripts/package.ps1
```diff
@@ -0,0 +1,54 @@
...
+if ($TAG | Select-String -Pattern 'dirty') {
+    $TAG = "dev"
+}
+
+if ($env:DRONE_TAG) {
+    $TAG = $env:DRONE_TAG
+}
+
+# Get release id as image tag suffix
+$HOST_RELEASE_ID = (Get-ItemProperty 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\' -ErrorAction Ignore).ReleaseId
```
luthermonson, line 29: this release id stuff doesnt work any more, it was an idea a while back for running docker in hyperv and doesnt work with how we have have drone hosts today. you can probably ditch all this logic and just have the else statement below. we've done this before in fleet

aiyengar2: dropped in latest commit. Let me know if the change looks good to you and I can squash

make.bat
```diff
@@ -18,7 +18,7 @@ goto :eof
 
 :.dapper
 if not exist .dapper.exe (
-    bitsadmin.exe /rawreturn /transfer dappwer-download https://releases.rancher.com/dapper/latest/dapper-Windows-x86_64.exe %~dp0\.dapper.exe
+    bitsadmin /rawreturn /transfer dappwer-download https://releases.rancher.com/dapper/latest/dapper-Windows-x86_64.exe %~dp0\.dapper.exe
```
nickgerace, line 21: the `exe` needed to be dropped?

scripts/release.ps1
```diff
@@ -1,5 +1,35 @@
...
+}
+
+docker push $pushTag
+if ($?) {
+    docker rmi "$pushTag-bak" | Out-Null
+    docker rmi $currentTag | Out-Null
+    Write-Host "$pushTag was PUSHED"
+} else {
+    docker tag "$pushTag-bak" $pushTag | Out-Null
+    docker rmi "$pushTag-bak" | Out-Null
```
nickgerace, line 33: What are the `bak` images for?

nickgerace (approved):
LGTM barring a few comments

make.bat
aiyengar2: Doesn't need to be dropped, but matches what we have in wmi_exporter-package https://github.com/rancher/wmi_exporter-package/blob/8e0c309b986cfbfb96ec1dba0a40df612f149f36/make.bat#L25

scripts/release.ps1
aiyengar2: copied from https://github.com/rancher/wmi_exporter-package/blob/8e0c309b986cfbfb96ec1dba0a40df612f149f36/scripts/windows/release.ps1#L18, but seems like they just create backup images in case the docker push fails (e.g. if I'm trying to push a local copy to the `rancher/` registry instead of `arvindiyengar/` by accident)

.drone.yml
```diff
@@ -8,37 +8,391 @@ platform:
...
+      ref:
+        - refs/heads/master
+        - refs/tags/*
+  - name: slack_notify
+    image: plugins/slack
     settings:
-      api_key:
-        from_secret: github_token
-      files:
-        - bin/wins.exe
```
luthermonson, line 31: do we want to miss out on publishing the binary on the release?

luthermonson: oh never mind i finally found it above... that was buried...

package/windows/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG SERVERCORE_VERSION
+
+FROM mcr.microsoft.com/powershell:nanoserver-${SERVERCORE_VERSION}
```
luthermonson, line 3: if all youre doing in is `wins.exe --version` you can skip using pwsh or servercore... just use `mcr.microsoft.com/windows/nanoserver:` and set the `CMD ["wins.exe", "--version"]`. can have 1 docker file then and bypass the powershell wrapper. agree?

.drone.yml
```diff
@@ -8,37 +8,391 @@ platform:
   version: 1809
 
 steps:
-  - name: build
-    pull: default
-    image: maiwj/dapper:windows-1809
-    commands:
-      - dapper.exe -f Dockerfile.dapper ci
+- name: build
```
luthermonson, line 11: we might want to talk about what youre trying to accomplish but i dont think your images outside of 1809 will have binaries if you dont build them on each build host

aiyengar2: Ah good catch! I didn't realize I omitted the build step on each host. Adding to the next commit

package/windows/Dockerfile
aiyengar2: We need Powershell since the actual script that will be run on this image is [upgrade.ps1](https://github.com/rancher/wins/blob/8b4d2f4aaf0cd01f8dfc6544703dd449c0c3176e/charts/rancher-wins-upgrader/scripts/upgrade.ps1) as this is meant to be a Docker image that will be used by the upgrade chart.

As discussed offline, I'll be modifying the Dockerfile to use servercore and only have 1 docker image for 20H2 + older versions.

aiyengar2: Addressed in latest commit.

luthermonson (approved):

PR #27, opened 2021-03-12
Fix Drone errors

aiyengar2:


.drone.yml
```diff
@@ -20,7 +20,7 @@ steps:
 - name: stage-binaries
   image: rancher/dapper:v0.5.4
   commands:
-    - "Get-ChildItem -Path ./bin; Copy-Item -Force -Path ./bin/wins.exe -Destination ./package/; Get-ChildItem -Path ./package/"
+    - "Get-ChildItem -Path ./bin; Copy-Item -Force -Path ./bin/wins.exe -Destination ./package/windows/; Get-ChildItem -Path ./package/windows"
```
aiyengar2, line 23: Needed to move the wins.exe into ./package/windows since that's the working directory fo the Dockerfile that references it in COPY

Since this is required at every build stage, the same exact code has be duplicated across all build pipelines.

package/windows/Dockerfile
```diff
@@ -3,11 +3,12 @@ ARG SERVERCORE_VERSION
 FROM mcr.microsoft.com/windows/servercore:${SERVERCORE_VERSION} as download
 ENV ARCH=amd64
 
+SHELL ["powershell", "-NoLogo", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
```
aiyengar2, line 6: Need this to run the pwsh symlink logic

package/windows/Dockerfile
```diff
@@ -3,11 +3,12 @@ ARG SERVERCORE_VERSION
...
+
 # Create a symbolic link pwsh.exe that points to powershell.exe for consistency
 RUN New-Item -ItemType SymbolicLink -Target "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe" -Path "C:\Windows\System32\WindowsPowerShell\v1.0\pwsh.exe"
 
-COPY ./bin/wins.exe /Windows/
-COPY entry.ps1 entry.ps1
+COPY ./wins.exe /Windows/
```
aiyengar2, line 11: We place wins.exe in `./package/windows/wins.exe`, which is why we refer to it as `./wins.exe` now

scripts/ci.ps1
```diff
@@ -6,5 +6,4 @@ Import-Module -WarningAction Ignore -Name "$PSScriptRoot\utils.psm1"
 
 Invoke-Script -File "$PSScriptRoot\test.ps1"
 Invoke-Script -File "$PSScriptRoot\build.ps1"
-
-Invoke-Expression -Command "$PSScriptRoot\package.ps1"
+Invoke-Script -File "$PSScriptRoot\package.ps1"
```
aiyengar2, line 9: nit change, can be reverted. Don't think it has any consequence but changed this to be consistent.

scripts/entry.ps1
```diff
@@ -13,8 +13,8 @@ $SCRIPT_PATH = ("{0}\{1}.ps1" -f $PSScriptRoot, $Args[0])
 if (Test-Path $SCRIPT_PATH -ErrorAction Ignore)
 {
     Invoke-Expression -Command $SCRIPT_PATH
-    if (-not $?) {
-        exit 255
+    if ($LASTEXITCODE -ne 0) {
```
aiyengar2, line 16: Required so that Docker build failures actually cause a failure on PR workflows; before, these errors were squashed and never returned.

scripts/package.ps1
```diff
@@ -41,4 +45,8 @@ docker build `
     -t $IMAGE `
     -f Dockerfile .
 
+if ($LASTEXITCODE -ne 0) {
```
aiyengar2, line 48: Required so that Docker build failures actually cause a failure on PR workflows; before, these errors were squashed and never returned.

scripts/package.ps1
```diff
@@ -5,6 +5,10 @@ Invoke-Expression -Command "$PSScriptRoot\version.ps1"
 
 $DIR_PATH = Split-Path -Parent $MyInvocation.MyCommand.Definition
 $SRC_PATH = (Resolve-Path "$DIR_PATH\..").Path
+
+# Reference binary in ./bin/wins.exe
+Copy-Item -Force -Path $SRC_PATH\bin\wins.exe -Destination $SRC_PATH\package\windows | Out-Null
```
aiyengar2, line 10: Identical to stage-binary logic, but we need to do this in `ci` since this runs within the Dapper container sandbox and doesn't make it out into the host.

luthermonson (approved):

PR #35, opened 2021-04-27
Leveraging the global for rkeWindowsPrefixPath.

phillipsj:
This PR includes leveraging the global.cattle.rkeWindowsPrefixPath global value to create the paths for the chart and powershell scripts.  This fixes the bug [here](https://github.com/rancher/rancher/issues/32194).


charts/rancher-wins-upgrader/scripts/upgrade.ps1
```diff
@@ -47,9 +47,15 @@ $winsUpgradeFilename = Split-Path -Path $winsUpgradePath -Leaf
 Create-Directory -Path $winsUpgradeDir
 Transfer-File -Src "c:\Windows\wins.exe" -Dst $winsUpgradePath
 
-Create-Directory -Path "c:\host\etc\rancher\wins"
-Transfer-File -Src $winsUpgradePath -Dst "c:\host\etc\rancher\wins\$winsUpgradeFilename"
-Transfer-File -Src "c:\scripts\config" -Dst "c:\host\etc\rancher\wins\config"
+if ($env:WINDOWS_HOST_PATH) {
```
nickgerace, line 50: Comment for most of the PR: will this variable always end with a path separator (`\`)?

charts/rancher-wins-upgrader/templates/_helpers.tpl
```diff
@@ -6,6 +6,14 @@
 {{- end -}}
 {{- end -}}
 
+{{- define "windows_prefix_path" -}}
+{{ ternary "C:/" .Values.global.cattle.rkeWindowsPrefixPath "" }}
+{{- end -}}
+
+{{- define "windows_host_path" -}}
+{{ include "windows_prefix_path" . | replace "C:/" "C:/host/" }}
```
nickgerace, line 14: I have two questions here...

1. We always prepend with `/host`?
2. What if `C:/` is not provided?

charts/rancher-wins-upgrader/templates/daemonset.yaml
```diff
@@ -37,7 +37,7 @@ spec:
         - name: wins-pipe
           mountPath: \\.\pipe\rancher_wins
         - name: wins
-          mountPath: c:/host/etc/rancher/wins
+          mountPath: {{ include "windows_host_path" $ }}etc/rancher/wins
```
nickgerace, line 40: This could be `winsUpgrader.winsHostPath` from above. Is this intentional?

charts/rancher-wins-upgrader/scripts/upgrade.ps1
phillipsj: Yep, I am setting it so it does. I create that in line 49 of daemonset.yaml and the function that is called is in the _helpers.tpl.

charts/rancher-wins-upgrader/templates/daemonset.yaml
phillipsj: I just kept it the same and substituted the one part of the path that needed to be different. Happy to change it to use the winsHostPath.

charts/rancher-wins-upgrader/scripts/upgrade.ps1
nickgerace: Regardless of user input, it'll always end with the path separator?

charts/rancher-wins-upgrader/templates/daemonset.yaml
nickgerace: I think that would be a good change :)

charts/rancher-wins-upgrader/templates/_helpers.tpl
phillipsj: Do we support having this installed on a different drive at all? Have we tested using a different root drive? I did this in the node cleanup and given that it is all based on windows specific values I just assumed. Good call out.

charts/rancher-wins-upgrader/scripts/upgrade.ps1
phillipsj: Hmm, good point. I pull what is placed in the config so there isn't user input. It may be potentially be left off.

charts/rancher-wins-upgrader/templates/_helpers.tpl
nickgerace: We should force the `C:/` drive. I am more concerned about handling whatever the user puts in `windows_prefix_path` :)

charts/rancher-wins-upgrader/scripts/upgrade.ps1
phillipsj: We control the windows hostpath and that will never have user input since that is inside our container.

charts/rancher-wins-upgrader/templates/_helpers.tpl
phillipsj: @nickgerace  I updated it and did a force push so you may have to dig for that change.

nickgerace (approved):

aiyengar2 (changes requested):

charts/rancher-wins-upgrader/values.yaml
```diff
@@ -7,10 +7,7 @@
 global:
   cattle:
     systemDefaultRegistry: ""
-
-## The path that prefixes all of your wins processPaths, used in the template for each config
-##
-prefixPath: c:\\
+    rkeWindowsPathPrefix: ""
```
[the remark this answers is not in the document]
aiyengar2: Seems like this was the intention based on https://github.com/rancher/dashboard/pull/2841.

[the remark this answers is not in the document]
phillipsj: Yes that was fixed and merged here:  https://github.com/rancher/dashboard/pull/2841/commits/85d1ee9c549011ea1b4a45a454ff9ed209a6f315

aiyengar2:
@phillipsj can you introduce the os annotation here as well, re: https://github.com/rancher/charts/blob/2620871639f461c0a8dc4786a9ee286b36a71b9d/packages/rancher-wins-upgrader/generated-changes/patch/values.yaml.patch#L10

charts/rancher-wins-upgrader/templates/_helpers.tpl
```diff
@@ -30,7 +30,7 @@ provider: kubernetes
 {{- end -}}
 
 {{- define "winsUpgrader.winsHostPath" -}}
-{{ .Values.prefixPath | replace "\\\\" "/" }}etc/rancher/wins
+{{ .Values.global.cattle.rkeWindowsPrefixPath | replace "\\\\" "/" }}etc/rancher/wins
```
aiyengar2, line 33: Do you think we need this replace statement anymore? Or should we just drop this and add a comment on rkeWindowsPathPrefix that it needs to use forward slashes?

charts/rancher-wins-upgrader/values.yaml
```diff
@@ -26,18 +23,18 @@ winsConfigs:
...
-        - {{ .Values.prefixPath }}etc\kubernetes\bin\kubelet.exe
-        - {{ .Values.prefixPath }}etc\nginx\nginx.exe
-        - {{ .Values.prefixPath }}opt\bin\flanneld.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/rancher/wins/wins-upgrade.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/windows-exporter/windows-exporter.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/wmi-exporter/wmi-exporter.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/kubernetes/bin/kube-proxy.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/kubernetes/bin/kubelet.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}etc/nginx/nginx.exe
+        - {{ .Values.global.cattle.rkeWindowsPrefixPath }}opt/bin/flanneld.exe
```
aiyengar2, line 32: Doesn't flipping the slashes cause an issue with the wins config? this would be exactly what is deposited in the wins config, so I had initially designed this to match what we put in our bootstrap script in https://github.com/rancher/rancher/blob/29e8af8818e56dc35e1bac239685803f62baa4ba/package/windows/bootstrap.ps1#L280-L293.

If we want to do forward slashes, we may need to modify the tpl call here to do ` {{ tpl .config $ | replace "/" "\\" indent 4 }}`:

https://github.com/rancher/wins/blob/72fac8616d77b23e7ce9c325622718e6a9857679/charts/rancher-wins-upgrader/templates/configmap.yaml#L10

charts/rancher-wins-upgrader/values.yaml
```diff
@@ -59,4 +56,4 @@ masquerade:
   # wmi_exporter is the only default whitelisted process that may or may not be run on the host, since
   # it is only ever deployed if the Windows cluster is also using Prometheus-based Windows monitoring (e.g. Rancher Monitoring V1)
   # All of the other default whitelisted processes are required for the Kubernetes cluster to operate
-  as: c:\\etc\wmi-exporter\wmi-exporter.exe
\ No newline at end of file
+  as: {{ .Values.global.cattle.rkeWindowsPrefixPath  }}etc/wmi-exporter/wmi-exporter.exe
```
aiyengar2, line 59: To allow us to pass in a Helm template, we'll need to add the tpl call to this between the required call and the first replace call:

https://github.com/rancher/wins/blob/72fac8616d77b23e7ce9c325622718e6a9857679/charts/rancher-wins-upgrader/templates/_helpers.tpl#L37

aiyengar2 (changes requested):

charts/rancher-wins-upgrader/templates/_helpers.tpl
phillipsj: You probably know the best option for that. I am happy to make whatever changes you think are best.

aiyengar2 (approved):
LGTM

nickgerace (approved):
MERGE TIME

PR #40, opened 2021-05-04
Ignore case when checking process path whitelist

aiyengar2:


tests/integration/process_test.ps1
```diff
@@ -35,8 +35,8 @@ Describe "process" {
         $config = @{
             whiteList = @{
                 processPaths = @(
-                    "c:\otherpath"
-                    "c:\etc\nginx\nginx.exe"
+                    "C:\otherpath"
+                    "C:\etc\nginx\nginx.exe"
```
aiyengar2, line 39: dummy hack to make the testing suite actually check whether this fix worked.

luthermonson (approved):

PR #42, opened 2021-05-05
Use provided winsUpgradePath as is for masquerade

aiyengar2:


charts/rancher-wins-upgrader/scripts/upgrade.ps1
```diff
@@ -42,7 +42,7 @@ if ($env:CATTLE_PREFIX_PATH) {
 }
 $winsUpgradePath = $('{0}etc\rancher\wins\wins-upgrade.exe' -f $prefixPath)
 if ($env:WINS_UPGRADE_PATH) {
-    $winsUpgradePath = $('{0}{1}' -f $prefixPath -f $env:WINS_UPGRADE_PATH)
+    $winsUpgradePath = $env:WINS_UPGRADE_PATH
```
aiyengar2, line 45: WINS_UPGRADE_PATH already contains prefixPath:

https://github.com/rancher/wins/blob/d80acc87e9450e9b1ab1d78aad0aa5e138ac57a6/charts/rancher-wins-upgrader/templates/daemonset.yaml#L43-L51

https://github.com/rancher/wins/blob/d80acc87e9450e9b1ab1d78aad0aa5e138ac57a6/charts/rancher-wins-upgrader/templates/_helpers.tpl#L45-L47

https://github.com/rancher/wins/blob/d80acc87e9450e9b1ab1d78aad0aa5e138ac57a6/charts/rancher-wins-upgrader/values.yaml#L54-L60

luthermonson: Should there be a default just in case?

luthermonson: Never mind it's right there two lines up

luthermonson (approved):

PR #43, opened 2021-05-06
adding powershell.exe to default config

luthermonson:
https://github.com/rancher/rancher/issues/32304

### DIFF
--- a/charts/rancher-wins-upgrader/values.yaml
+++ b/charts/rancher-wins-upgrader/values.yaml
@@ -25,6 +25,7 @@ winsConfigs:
     whiteList:
       processPaths:
         - {{ default "c:\\" .Values.global.cattle.rkeWindowsPathPrefix | replace "/" "\\" }}etc\rancher\wins\wins-upgrade.exe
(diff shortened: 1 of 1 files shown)


aiyengar2 (commented):
We haven't been doing this in past PRs, but going forward (now that this chart has been officially released), we should bump the chart's minor version every time we make a change.

Can you add that to this PR? Otherwise, LGTM.

aiyengar2:
FYI @luthermonson once this PR is merged we also need to bump the commit here https://github.com/rancher/charts/blob/d036725e08670138c5d9730c5dd4e8c00dfa8c43/packages/rancher-wins-upgrader/package.yaml#L3-L5

We should then reset the packageVersion and releasedCandidateVersion to `01` (new convention, previously we were setting them to 00).

aiyengar2:
Also @luthermonson I think we have discussed this offline, but while this change is a nice-to-have in the wins-upgrader chart, I think it's important to call out that users may be using the older version of wins upgrader and may not have updated to the latest (or may not be using this chart at all). 

So if your changes for cleanup are dependent on this path being whitelisted, it needs to be able to handle the case that this path is not whitelisted.

PR #49, opened 2021-09-08
Removing old powershell implementation and using the one in the upgrader

phillipsj:
Most of the powershells package wasn't actually used outside of unit tests. The single place it was used was in the process management package.


pkg/apis/process_service_mgmt.go
```diff
@@ -13,8 +13,9 @@ import (
 	"syscall"
 	"unsafe"
 
+	"github.com/rancher/wins/pkg/powershell"
+
```
luthermonson, line 17: Nit goimports

luthermonson (approved):

PR #55, opened 2021-09-15
adding fossa step to 1809

luthermonson:
adds fossa scanning to the 1809 build process after the build

https://github.com/rancherlabs/eio/issues/504


rosskirkpat (approved):
LGTM

PR #56, opened 2021-10-01
Adding system agent support

phillipsj:
This is the basic work to get system agent running in Wins.


cmd/server/app/run.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/rancher/wins/pkg/sa"
+
```
luthermonson, line 8: goimports

rosskirkpat (approved):
LGTM

brandond (approved):

dweomer (approved):

cmd/server/app/run_service.go
```diff
@@ -2,6 +2,7 @@ package app
 
 import (
 	"context"
+	"github.com/rancher/wins/pkg/sa"
```
nickgerace, line 5: nit: We should move this dep

cmd/server/app/run_service.go
```diff
@@ -241,6 +244,11 @@ func (h *serviceHandler) Execute(_ []string, r <-chan svc.ChangeRequest, s chan<
 		h.errC <- h.srv.Serve(ctx)
 	}()
 
+    // start system agent
+	go func() {
+		h.errC <- h.agent.Run(ctx)
+	}()
```
nickgerace, line 250: nit: `fmt`

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type Agent struct {
+	cfg *config.AgentConfig
+}
+
+func (a *Agent) Run(ctx context.Context) error {
+
+	if a.cfg == nil {
+		return nil
```
nickgerace, line 22: Should we error out with a nil config instead? What's the real world scenario when this is ran with a nil config?

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+		var connInfo config.ConnectionInfo
+
+		if err := config.Parse(a.cfg.ConnectionInfoFile, &connInfo); err != nil {
+			logrus.Fatalf("Unable to parse connection info file %v", err)
+		}
+
+		k8splan.Watch(ctx, *applier, connInfo)
+	}
+
+	if a.cfg.LocalEnabled {
```
nickgerace, line 48: nit: `else if` might be better here for readability

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+	if a.cfg.RemoteEnabled {
+		logrus.Infof("Starting remote watch of plans")
+
+		var connInfo config.ConnectionInfo
+
+		if err := config.Parse(a.cfg.ConnectionInfoFile, &connInfo); err != nil {
+			logrus.Fatalf("Unable to parse connection info file %v", err)
+		}
+
+		k8splan.Watch(ctx, *applier, connInfo)
```
nickgerace, line 45: Does this watcher handle its own errors?

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+	imageUtil := image.NewUtility(a.cfg.ImagesDir, a.cfg.ImageCredentialProviderConfig, a.cfg.ImageCredentialProviderBinDir, a.cfg.AgentRegistriesFile)
+	applier := applyinator.NewApplyinator(a.cfg.WorkDir, a.cfg.PreserveWorkDir, a.cfg.AppliedPlanDir, imageUtil)
+
+	if a.cfg.RemoteEnabled {
+		logrus.Infof("Starting remote watch of plans")
+
+		var connInfo config.ConnectionInfo
+
+		if err := config.Parse(a.cfg.ConnectionInfoFile, &connInfo); err != nil {
+			logrus.Fatalf("Unable to parse connection info file %v", err)
```
nickgerace, line 42: Why not return this? We need to exit?

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	return nil
+}
+
+func New(cfg *config.AgentConfig) (*Agent, error) {
+	return &Agent{
+		cfg: cfg,
+	}, nil
+}
```
nickgerace, line 60: This function doesn't seem necessary to me. We don't need to future proof an error statement IMO.

nickgerace (changes requested):

pkg/sa/agent.go
phillipsj: If the agent config isn't set, aka nil, then it's a no op. There is a decision record for t design.

pkg/sa/agent.go
phillipsj: It's how it is used upstream in system agent.

pkg/sa/agent.go
phillipsj: This is how it's used upstream and it does handle errors.

docs/adrs/0002-add-system-agent-support-to-wins.md
```diff
@@ -0,0 +1,19 @@
...
+
+Accepted
+
+## Context
+
+The [System Agent]() is used to install and upgrade RKE2 using plans. Support is needed to provide this ability for Windows. 
+
+## Decision
+
+System Agent functionality needs some capabilities of Wins to achieve its goal. This would require that both the system agent service and wins service to be running on a system. Alternatively, we could consume the system agent as a package in the wins service reducing the need to having both installed in a system.
```
aiyengar2, line 15: nit: Seems like this should be part of the Context, not the decision. The decision is just the last sentence right?

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+func (a *Agent) Run(ctx context.Context) error {
+
+	if a.cfg == nil {
+		return nil
+	}
+
+	logrus.Infof("Rancher System Agent version %s is starting", version.FriendlyVersion())
+
+	if !a.cfg.LocalEnabled && !a.cfg.RemoteEnabled {
+		logrus.Fatalf("Local and remote were both not enabled. Exiting, as one must be enabled.")
```
aiyengar2, line 28: Should the behavior here also be to return `nil`, based on https://github.com/rancher/wins/pull/56/files#r721730809? If so, should we move this before the `logrus` line?

pkg/sa/agent.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	logrus.Infof("Using directory %s for work", a.cfg.WorkDir)
```
aiyengar2, line 31: nit: 
```suggestion
	logrus.Infof("Setting %s as the working directory", a.cfg.WorkDir)
```

pkg/sa/agent.go
aiyengar2: should we add a colon before the `%v`?

pkg/sa/agent.go
aiyengar2: Agreed. I don't think this is necessary.

aiyengar2 (changes requested):

pkg/sa/agent.go
nickgerace: My mistake, I did not realize you could use both at once.

pkg/sa/agent.go
nickgerace: ^ that's a `nit` if I've ever seen one 
image

nickgerace (approved):

aiyengar2 (approved):
LGTM

cmd/server/app/run.go
```diff
@@ -104,7 +105,12 @@ func _runAction(cliCtx *cli.Context) error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to create server")
 	}
-	err = runService(ctx, server)
+	agent := systemagent.New(cfg.SystemAgent)
+	if err != nil {
```
briandowns, line 109: What error is being checked here?

cmd/server/app/run_service.go
```diff
@@ -241,6 +244,11 @@ func (h *serviceHandler) Execute(_ []string, r <-chan svc.ChangeRequest, s chan<
 		h.errC <- h.srv.Serve(ctx)
 	}()
 
+	// start system agent
+	go func() {
+		h.errC <- h.agent.Run(ctx)
```
briandowns, line 249: Should we be mixing the use of this channel wtih agent and server?

phillipsj: That channel is for the windows service, so both "sub processes" should be reporting into that channel.

briandowns (approved):

PR #66, opened 2021-11-22
ADR for CSI Proxy

phillipsj:
This PR is for the proposed architecture decision for CSI Proxy and to enable the discussion.

### DIFF
--- a/docs/adrs/0004-csi-proxy-and-configuration.md
+++ b/docs/adrs/0004-csi-proxy-and-configuration.md
@@ -1,0 +1,27 @@
+# 4. CSI Proxy and Configuration
+
+Date: 2021-11-22
+
+## Status
+
+Approved
+
+## Context
+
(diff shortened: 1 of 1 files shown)


rosskirkpat (approved):
This is the start of an exciting step forward of both using ADRs and enabling csi-proxy functionality

docs/adrs/0004-csi-proxy-and-configuration.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Decision
+
+After some discussion, it was decided that Wins would be responsible for downloading the specified version of CSI Proxy from the current upstream location based on the version that is specified. The ability to override the upstream URL will be provided. Once the binary is available Wins will be responsible for creating the Windows Service for CSI Proxy and ensuring that it is running. Wins will also have the ability to manage the lifecycle of the CSI Proxy service since configuration changes will need to be picked up from Wins. The proposed configuration is below:
+
+```yaml
+csi-proxy:
+  url: <location of the binaries>
+  version: <version>
+  managed: <bool>
```
luthermonson, line 21: this is a bool to manage the service right? perhaps we only manage it if the `csi-proxy` config is here we just manage it any way... why would someone add this config and not expect wins to manage it?

luthermonson (approved):

PR #71, opened 2021-12-13
Copying the wins.exe to the location expected by monitoring chart.

phillipsj:
This just copies the wins.exe as part of the instal.ps1 to guarantee that Wins in the location expected in the charts. This was handled by the Rancher-agent prior to RKE2.

* https://github.com/rancher/rancher/issues/35851

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -536,6 +536,14 @@ systemagent:
         }
     }
 
+    function Copy-WinsForCharts() {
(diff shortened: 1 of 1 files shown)


rosskirkpat (approved):

phillipsj:
This passed deployment checks and the Wins exporter issue was addressed
35851-cluster-management
.

PR #73, opened 2021-12-14
feat: support to mount volume from container to host

thxCode:
what we do in this pr:
- implement a new clientset for `volume`
- support mounting the volume of the container instead of transferring some artifacts outside, it's used to deal with some configmap/secret mount points.

example:
1. deploy a workload with some volume mountings.
```
          volumeMounts:
            - name: addon-token
              mountPath: 'c:/var/addon'
            - name: wins-pipe
              mountPath: \\.\pipe\rancher_wins
      volumes:
        - name: addon-token
          secret:
            secretName: addon.network.token
            items:
              - key: addon.token.config
                path: token-config
         - name: wins-pipe
          hostPath:
            path: \\.\pipe\rancher_wins
            type: ''
```
2. without `volume mount`, transfer the file to the host previously, but it cannot deal with the expired addon token.
```
Transfer-File -Src "c:\var\addon\token-config" -Dst "c:\host\var\addon\token-config"
```
3. with `volume mount`, declare the current container selectors and the path to mount.
```
$vol_selectors = @(
  "io.kubernetes.pod.namespace=${env_pod_namespace}"
  "io.kubernetes.pod.name=${env_pod_name}"
  "io.kubernetes.container.name=${env_container_name}"
  "io.kubernetes.docker.type=container"
)
$vol_paths = @(
  "c:\var\addon"
)
wins cli vol mount --selectors="$($vol_selectors -join ' ')" --paths="$($vol_paths -join ' ')"
```


luthermonson:
@thxCode you sure you want this? we just implmented csi proxy support in wins and shipped it in v0.1.2. and by the way hi frank :) good to see your face again

thxCode:
hey @luthermonson , i think it's awesome that v0.1.2 can help us to install the csi proxy, but the `wins cli volume mount` function is something like proxying the configuration projection from the container to host. 

like the above example, the `addon.network.token` secret can be refreshed interval, the hosting process cannot see the change of the configuration. btw, is it possible to implement this via csi proxy?

PR #85, opened 2022-03-02
Improving the curl switch handling.

phillipsj:
There was an issue with how the switch statements for checking curl results were being processed.

* https://github.com/rancher/windows/issues/151

image


rosskirkpat (approved):
lgtm

PR #86, opened 2022-03-04
pin v0.2.0 of system-agent

luthermonson:
a lot of changes went into system agent for rancher v2.6.4, pinning latest version of system-agent for feature parity.


rosskirkpat (approved):
lgtm

PR #89, opened 2022-03-07
Correcting issues related to proxy for PowerShell commands and regedit

phillipsj:
### Root cause

PowerShell 5.1, the default on Windows Server, doesn't support consuming the environment variables for proxy configuration. This is fixed in newer versions but want to be backported. Another issue occurred that wasn't saving these to the registry correctly for wins.
 
### What was fixed, or what changes have occurred

Switched to using curl instead of Invoke-RestMethod and made a change to how the registry key was getting edited/added.
 
Installation of a cluster from Rancher.
 
### What areas could experience regressions?

Installation of a Windows node using the command from Rancher.
 
### Are the repro steps accurate/minimal?

Configure a system using a proxy and set the agent vars in Rancher when creating the cluster. 

### Related Issues

* https://github.com/rancher/rancher/issues/36732

proxy-wins-installer

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -189,6 +189,8 @@ function Invoke-WinsInstaller {
     }
 
     function Set-Environment {
+        $env:CURL_CAFLAG = "--ssl-no-revoke"
+
         if (-Not $env:CATTLE_ROLE_CONTROLPLANE) {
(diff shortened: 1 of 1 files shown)


rosskirkpat (approved):
LGTM

PR #100, opened 2022-03-18
pin v1.20.0 of urfave for a args parsing error with double quotes

luthermonson:
https://github.com/rancher/wins/pull/96


rosskirkpat (approved):
lgtm

PR #116, opened 2022-05-06
fix tlsConfig for csi proxy and install script

rosskirkpat:
This fixes an issue where the install script was over-riding existing wins config as well as sets the default csiProxy http transport to be insecure.

### Summary
Fixes a new bug found while testing https://github.com/rancher/rancher/issues/36732 and https://github.com/rancher/rancher/issues/36574

- do not override the wins config when using the wins install script
- ensure that the default certFilePath in wins config is empty for non-rancher users
- set certFilePath as part of the wins install script
- call ValidateTLSConfig as part of wins config validation
- default to using an insecure http transport to match system-agent functionality

### Technical notes summary

in csi.go, we are now specifying an http.Transport struct as it is the only way to mimic a `--insecure` that is used in the system-agent scripts.

I tested the install script on a windows machine

I don't see any areas of possible regressions.

### Screenshot/Video

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -468,7 +468,7 @@ white_list:
   proxyPorts:
     - 9796
 "@
-        Set-Content -Path $env:CATTLE_AGENT_CONFIG_DIR/config -Value $winsConfig
+        Add-Content -Path $env:CATTLE_AGENT_CONFIG_DIR/config -Value $winsConfig
 
         $agentConfig = 
         @"
@@ -477,6 +477,8 @@ systemagent:
   appliedPlanDirectory: $($env:CATTLE_AGENT_VAR_DIR)/applied
   remoteEnabled: $($env:CATTLE_REMOTE_ENABLED)
   preserveWorkDirectory: $($env:CATTLE_PRESERVE_WORKDIR)
+tls-config:
+  certFilePath: C:/etc/rancher/wins/ranchercert
 "@
         Add-Content -Path $env:CATTLE_AGENT_CONFIG_DIR/config -Value $agentConfig
         if ($env:CATTLE_REMOTE_ENABLED -eq "true") {
--- a/install.ps1
+++ b/install.ps1
@@ -472,13 +472,13 @@ white_list:
 
         $agentConfig = 
         @"
+tls-config:
+  certFilePath: C:/etc/rancher/wins/ranchercert
 systemagent:
   workDirectory: $($env:CATTLE_AGENT_VAR_DIR)/work
   appliedPlanDirectory: $($env:CATTLE_AGENT_VAR_DIR)/applied
   remoteEnabled: $($env:CATTLE_REMOTE_ENABLED)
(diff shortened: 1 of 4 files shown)


phillipsj (approved):
LGTM!

phillipsj (approved):
LGTM!

pkg/csiproxy/csi.go
```diff
@@ -125,13 +126,24 @@ func (p *Proxy) download() error {
...
+	}
+
 	resp, err := client.Get(fmt.Sprintf(p.cfg.URL, p.cfg.Version))
 	if err != nil {
 		return err
 	}
 
 	defer func(Body io.ReadCloser) {
 		_ = Body.Close()
+		transport.CloseIdleConnections()
```
luthermonson, line 146: you should do a second defer up above as the client.Get could error and you might leak connections if you only have it in this defer

install.ps1
```diff
@@ -468,10 +468,12 @@ white_list:
   proxyPorts:
     - 9796
 "@
-        Set-Content -Path $env:CATTLE_AGENT_CONFIG_DIR/config -Value $winsConfig
```
luthermonson, line 471: didnt we have this on purpose so it resets the contents of the config on a reboot so we dont append forever?

phillipsj: @luthermonson why would this run on a reboot? This just run once during registration and the other stuff does an append also.

luthermonson: we have the upgrade process that will be calling this i guess is what i was referring too

luthermonson: https://github.com/rancher/system-agent/blob/main/package/suc/run.sh#L48

pkg/csiproxy/csi.go
```diff
@@ -125,6 +126,18 @@ func (p *Proxy) download() error {
...
+	// default to insecure which matches system-agent functionality
+	transport := &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
+
+	if !*p.cfg.Insecure && p.cfg.CertFilePath != "" {
+		transport.TLSClientConfig.InsecureSkipVerify = false
+	}
+
+	client.Transport = transport
+
+	defer client.CloseIdleConnections()
```
luthermonson, line 139: much gooder~

phillipsj (approved):

install.ps1
rosskirkpat: There was an (undocumented) change in behavior as we only use to append to the wins config file and now we replace any existing wins config file. I am setting the behavior back to appending only as it makes it impossible to test custom wins configuration and may impact our community end users. In my case, I was trying to test my changes while writing this PR by manually creating a wins config with `tls-config` and it would be wiped out when the install script was run.

luthermonson (approved):

rosskirkpat:
https://github.com/rancher/windows/issues/161

PR #142, opened 2022-09-15
Revert EventLog support for SUC script back to Write-Host

Oats87:
### Summary
Fixes https://github.com/rancher/rancher/issues/38896 and https://github.com/rancher/rancher/issues/38915

A change was made in PR https://github.com/rancher/wins/pull/141 that changed logs to being written to both the EventLog and host. The issue with this is that the EventLog didn't exist within the container, and rather than attempt to create the containerized EventLog, it's probably easier to just write logs to the host. 

This PR retains the increased log verbosity but does not attempt to write to the EventLog anymore.

We should make sure the SUC image still works.

The SUC image.

### DIFF
--- a/suc/run.ps1
+++ b/suc/run.ps1
@@ -24,12 +24,12 @@ function New-Directory {
     if (Test-Path -Path $Path) {
         if (-not (Test-Path -Path $Path -PathType Container)) {
             # clean the same path file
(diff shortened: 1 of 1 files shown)


jakefhyde (approved):
LGTM, got the feeling I'm going to need to set up a windows vm real soon...

slickwarren (approved):

PR #158, opened 2024-01-19
Run Integration tests on GHA

pjbgf:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,46 @@
...
+    
+    - name: Install Dependencies
+      run: |
+        go install github.com/magefile/mage@latest
+        go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+
+    - name: Run E2E tests
+      shell: pwsh
+      run: |
+        Install-Module -Name DockerMsftProvider -Force
```
HarrisonWAffel, line 44: Are the `Install-Module` and `Import-Module` statements required for this to work? My understanding was that both the windows 2019 and 2022 github runners came preconfigured with docker ([2019 details](https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md#tools), [2022 details](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#tools)). Also, the `DockerMsftProvider` [module is deprecated](https://github.com/OneGet/MicrosoftDockerProvider?tab=readme-ov-file) (But for whatever reason still seems to work).

tests/integration/process_test.ps1
```diff
@@ -31,57 +31,57 @@ Describe "process" {
...
-        Execute-Binary -FilePath "docker.exe" -ArgumentList @("rm", "-f", "prc-run") -PassThru | Out-Null
-        {
-            # should not be abled to find processes
-            {
-                Get-Process -Name "rancher-wins-*" -ErrorAction Ignore
-            } | Judge -Reverse -Throw
-        } | Should Not Throw
-        Get-NetFirewallRule -PolicyStore ActiveStore -Name "rancher-wins-*-TCP-80" -ErrorAction Ignore | Should BeNullOrEmpty
-    }
+    # It "run" {
```
HarrisonWAffel, line 34: Is there a reason that this is commented out and not removed entirely? I would think testing that we can deploy the nginx exe as well as confirming that the exe works as expected would be a useful test case. Are we limited from running this because we are moving towards GHA?

HarrisonWAffel (commented):
A few comments, but generally looks good!

tests/integration/process_test.ps1
HarrisonWAffel: Okay, sounds good!

.github/workflows/test.yaml
pjbgf: You are absolutely right, both 2019 and 2022 GHA runners come with Docker pre-installed.

The reason why I added this `Install-Module` and `Import-Module` was to be able to get [hns_test.ps1](https://github.com/rancher/wins/blob/main/tests/integration/hns_test.ps1) to run. Without this it fails with `Get-HnsNetwork` is to recognised IIRC.

HarrisonWAffel: Interesting, I thought the hns utility commands would be bundled together with docker support. I guess leaving the `DockerMsftProvider` line can't hurt, if it fails to install docker then we will just use what is already installed. The `Import-Module` line makes more sense now

HarrisonWAffel (approved):
Looks good, but I would like to wait for @aiyengar2 to review before we merge

.github/workflows/test.yaml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    # Environment requirements:
+    # https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment
+    - name: Test docker
+      run: |
+        docker version
+        docker pull hello-world
+        docker run hello-world
```
aiyengar2, line 26: don't think we need to run this step every time we run the workflow, since we know Docker will be part of the GH runner (based on Harrison's comment) we know this step should work.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 'stable'
```
aiyengar2, line 34: Shouldn't be necessary, it should already be installed as of https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md#language-and-runtime

aiyengar2 (changes requested):

aiyengar2 (approved):
Will approve for now so we can get tests running, will follow up on the changes I requested in later PRs. Thanks for the contribution!

.github/workflows/test.yaml
pjbgf: I removed this from a follow-up PR, and ended-up having issues with both windows versions:
https://github.com/rancher/wins/actions/runs/8094456813/job/22119187899?pr=159
https://github.com/rancher/wins/actions/runs/8094456813/job/22119187595?pr=159

PR #190, opened 2024-05-14
migrate to github actions

HarrisonWAffel:
### Summary
We need to migrate off of drone and onto github actions

Drop drone file, add release workflow to github actions

A successful release run can be seen here: https://github.com/HarrisonWAffel/wins/actions/runs/9038294743
The resulting docker images are here: https://hub.docker.com/r/harrisonwaffel/wins/tags


scripts/build-image.ps1
```diff
@@ -0,0 +1,46 @@
+<#
```
jakefhyde, line 1: any reason why this ins't a mage target? just curious, I have 0 context around wins

jakefhyde (approved):
Approving for sanity, but should probably have @aiyengar2 review for correctness

scripts/build-image.ps1
HarrisonWAffel: Wins doesn't have any docker utils that can be used by our mage files, and adding them may be a waste of time depending on how we want to standardize CI across all of the windows repos. So, yes it could be a mage target but for the sake of time and to avoid some of that work being undone in the future I opted for a simple powershell script instead.

aiyengar2 (approved):

PR #225, opened 2024-08-26
Improving the SUC image

HarrisonWAffel:
### Summary

The initial implementation of the `rancher/wins` image was intended to be used with the System Upgrade Controller via the Rancher System Agent Upgrader. However, this feature was never properly implemented within Rancher, and as a result Windows nodes cannot receive maintenance updates (binary upgrades, config updates, etc.) on existing Rancher clusters. Since this image was initially developed, new features have been introduced to Kubernetes on Windows which allows us to simplify the approach to upgrading Windows nodes once they have already been provisioned. 

This PR updates the `rancher/wins` SUC image to be compatible with Host Process pods, eliminating the need to mount the `rancher-wins` named pipe to perform privileged operations. As a part of this update, the related powershell scripts have been removed and replaced with a Go executable. This executable is capable of updating the `rancher-wins` config file based off of environment variables, managing the `rancher-wins` service, as well as being responsible for executing a simplified version of the `install.ps1` Powershell script with the goal of updating the `rancher2_connection_info.json` file. The logic surrounding management of the rancher-wins config file, as well as the powershell script for updating the `rancher2_connection_info.json` file are required for Rancher to offer support of the Agent TLS Mode feature on Windows nodes. 

Transitioning to a Go binary also allows us to significantly reduce the size of the `rancher/wins` image (~2GB -> ~200MB), by transitioning to a `nanoserver` base image. Now that we can run directly on the host, we do not need to repackage the Windows APIs and binaries that are included with the `servercore` image.

Additionally, the `rancher-wins-upgrader` chart has been removed. This chart has been deprecated for some time, and the chart annotations did not allow for it to be deployed onto Rancher versions post v2.6 (which is now EOL). 

This change will be used as a jumping off point for future improvements to the SUC image, with the ultimate goal of allowing automatic upgrades of `rancher-wins` and its related binaries


install.ps1
```diff
@@ -224,6 +227,12 @@ function Invoke-WinsInstaller {
             $env:CATTLE_AGENT_LOGLEVEL = $env:CATTLE_AGENT_LOGLEVEL.ToLower()
         }
 
+        if ((-Not $env:STRICT_VERIFY) -and ($StrictTlsVerification)) {
+            $env:STRICT_VERIFY = "true"
+        } elseif (-Not $env:STRICT_VERIFY){
+            $env:STRICT_VERIFY = "false"
+        }
```
jiaqiluo, line 234: Does it cover the case where the env var STRICT_VERIFY has been set to false and we want to change it to true by providing StrictTlsVerification?

magefiles/magefile.go
```diff
@@ -157,8 +160,8 @@ func Build() error {
 		return err
 	}
 
-	if len(files) != 5 {
-		return errors.New("[Build] a required build artifact is missing, exiting now \n")
+	if len(files) != requiredFilesCount {
+		return errors.New(fmt.Sprintf("[Build] a required build artifact is missing, expected %d artifacts and only got %d, exiting now \n", requiredFilesCount, len(files)))
```
jiaqiluo, line 164: Can it be simplified by using `fmt.Errorf()`?

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/urfave/cli/v2"
+)
+
+func Run(_ *cli.Context) error {
+	var errs []error
+	refreshService := false
+
+	initialState, err := service.BuildInitialState()
+	if err != nil {
+		return fmt.Errorf("could not build initial state for rancher-wins: %v", err)
```
jiaqiluo, line 19: The same suggestion to all occurrences: 
```suggestion
		return fmt.Errorf("could not build initial state for rancher-wins: %w", err)
```

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	logrus.Infof("successfully updated connection info")
+	if o != "" {
+		logrus.Infof(" Script output:\n%s", o)
+	}
+
+	// update the config using env vars
+	restartServiceDueToConfigChange, err := service.UpdateConfigFromEnvVars("")
+	refreshService = restartServiceDueToConfigChange
```
jiaqiluo, line 37:
```suggestion
	refreshService, err = service.UpdateConfigFromEnvVars("")
```

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if refreshService && (errs == nil || len(errs) == 0) {
```
jiaqiluo, line 42: `errs` has been defined as `[]error` so it could not be nil.

suc/pkg/rancher/connectionInfo.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func UpdateConnectionInformation() (string, error) {
+	connInfoScriptName := "/hpc/update-connection-info.ps1"
```
jiaqiluo, line 13: Can we make connInfoScriptName a const outside of the function?

suc/pkg/rancher/connectionInfo.go
```diff
@@ -0,0 +1,28 @@
...
+	_, err := os.Stat(connInfoScriptName)
+	if errors.Is(err, os.ErrNotExist) {
+		logrus.Errorf("Could not find update-connection-info.ps1, will not attempt to update Rancher connection information.")
```
jiaqiluo, line 16:
```suggestion
		logrus.Errorf("Could not find %s, will not attempt to update Rancher connection information.", connInfoScriptName)
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// UpdateConfigFromEnvVars is responsible for updating the rancher-wins config file
+// based off of the presence of particular environment variables. The path parameter is used
+// to specify the location of the config file on the host. If path is left empty, the defaultConfigFile
+// constant is used. The config file will only be updated if a given environment variable is present, and its
+// value does not equal the currently set value in the config file. UpdateConfigFromEnvVars returns a boolean
+// indicating if the config file has been updated and any errors encountered.
+func UpdateConfigFromEnvVars(path string) (bool, error) {
+	logrus.Infof("Loading config from host")
```
jiaqiluo, line 96:
```suggestion
	logrus.Info("Loading config from host")
```

suc/update-connection-info.ps1
```diff
@@ -0,0 +1,172 @@
...
+                Write-LogInfo "Successfully tested Rancher connection."
+                $env:RANCHER_SUCCESS = $true
+                $retries = 99
+                break
+            }
+            default {
+                Write-LogError "$responseCode received while testing Rancher connection. Sleeping for 5 seconds and trying again."
+                Start-Sleep -Seconds 5
+                $retries++
+                continue
```
jiaqiluo, line 85: The same suggestion to other similar while loop in other functions:
Could we log an addition error message to indicate the test fails because hitting the retries limit (5)?

jiaqiluo (changes requested):

cmd/server/config/config.go
```diff
@@ -122,6 +92,19 @@ func LoadConfig(path string, v *Config) error {
...
+	if v == nil {
+		return fmt.Errorf("config cannot be nil")
+	}
+
+	yml, err := yaml.Marshal(v)
+	if err != nil {
+		return fmt.Errorf("could not marshal provided config")
+	}
+
+	return os.WriteFile(path, yml, os.ModePerm)
```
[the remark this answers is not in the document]
HarrisonWAffel: Permission bits on Windows do not work properly for files created with Go, so `os.ModePerm` is no better than `600` - ultimately windows uses access control lists to limit who can use the file. There is work scheduled in the coming releases for fixing many of the ACLs on files created by wins and rancher, and I'd like to handle this file as a part of that work instead of in this PR

suc/pkg/service/service.go
```diff
@@ -0,0 +1,143 @@
...
+
+func waitForServiceState(s *mgr.Service, desiredState svc.State, delayInSeconds time.Duration, maxAttempts int) error {
+	transitionedSuccessfully := false
+	var err error
+	var state svc.State
+
+	for i := 0; i < maxAttempts; i++ {
+		state, err = getServiceState(s)
+		if err != nil {
+			return fmt.Errorf("failed to query service %s: %v", s.Name, err)
```
jakefhyde, line 111:
```suggestion
			return fmt.Errorf("failed to query service %s: %w", s.Name, err)
```

suc/pkg/service/service.go
```diff
@@ -0,0 +1,143 @@
...
+		return err
+	}
+
+	return waitForServiceState(s, svc.Running, 5, 5)
+}
+
+func openService(name string) (*mgr.Service, error) {
+	svcMgr, err := mgr.Connect()
+	if err != nil {
+		return nil, err
```
jakefhyde, line 88: IMO it would be a good idea to wrap all errors.

suc/pkg/service/config_test.go
```diff
@@ -0,0 +1,149 @@
...
+				def.Debug = true
+				def.AgentStrictTLSMode = true
+				return def
+			},
+			updateExpected: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
```
jakefhyde, line 118:
```suggestion
  t.Parallel()
  
	for _, tc := range tests {
	  tc := tc
		t.Run(tc.name, func(t *testing.T) {
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+			logrus.Errorf("Error encountered while pasring %s, field will not be updated: %v", AgentStringTLSEnvVar.getSimple(), err)
+		} else if cfg.AgentStrictTLSMode != tlsMode {
+			cfg.AgentStrictTLSMode = tlsMode
+			configNeedsUpdate = true
+		}
+	}
+
+	// If we haven't made any changes there is no reason to update the config file
+	if configNeedsUpdate {
+		logrus.Infof("Detected a change in configuration, updating config file")
```
jakefhyde, line 127:
```suggestion
		logrus.Info("Detected a change in configuration, updating config file")
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+		err = saveConfig(cfg, path)
+		if err != nil {
+			return configNeedsUpdate, fmt.Errorf("failed to save config: %v", err)
+		}
+	} else {
+		logrus.Infof("Did not detect a change in configuration")
```
jakefhyde, line 133:
```suggestion
		logrus.Info("Did not detect a change in configuration")
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+var (
+	ConfigDirEnvVar      = envVar{varName: "wins_config_dir"}
+	DebugEnvVar          = envVar{varName: "debug"}
+	AgentStringTLSEnvVar = envVar{varName: "strict_verify"}
+)
+
+const (
+	defaultConfigFile = "C:/etc/rancher/wins/config"
```
jakefhyde, line 20: Why is `C` sometimes capitalized and sometimes not?

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+func (e *envVar) get() string {
+	return os.Getenv(e.name())
+}
+
+// getSimple returns the value of the environment variable
+// using os.Getenv. The simple name is passed to os.Getenv
+// (i.e. there is no 'CATTLE_WINS_' prefix)
+func (e *envVar) getSimple() string {
+	return os.Getenv(e.simpleName())
+}
```
jakefhyde, line 39: Is this necessary? I feel like we only are ever using the `CATTLE_WINS_` version or raw version in this file, but it doesn't appear like we are mixing them.

suc/pkg/service/config.go
```diff
@@ -0,0 +1,137 @@
...
+// a config file located on the host. If both are provided, the path parameter will take
+// precedence. If neither are provided, then the defaultConfigFile path is used.
+func loadConfig(path string) (*config.Config, error) {
+	configDirEnv := os.Getenv(ConfigDirEnvVar.get())
+	if path == "" && configDirEnv != "" {
+		path = configDirEnv
+	}
+	if path == "" {
+		path = defaultConfigFile
+	}
```
jakefhyde, line 63: IMO this could probably be it's own function like `getConfigDir()`

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+	return combineErrors(errs)
+}
+
+func combineErrors(errs []error) error {
+	var err error
+	for _, e := range errs {
+		err = errors.Join(err, e)
+	}
+	return err
+}
```
jakefhyde, line 64: Why can't we just `return errors.Join(errs)`?

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+
+	logrus.Infof("Updating rancher connection info")
+	o, err := rancher.UpdateConnectionInformation()
+	if err != nil {
+		logrus.Errorf("Could not update rancher connection information. Script output:")
+		logrus.Error(o)
+		return fmt.Errorf("error encountered while refreshing connection information: %v", err)
+	}
+
+	logrus.Infof("successfully updated connection info")
```
jakefhyde, line 30:
```suggestion
	logrus.Infof("Successfully updated connection info")
```

suc/pkg/cmd.go
```diff
@@ -0,0 +1,64 @@
...
+	if o != "" {
+		logrus.Infof(" Script output:\n%s", o)
```
jakefhyde, line 32: Is the space at the beginning intentional?

suc/pkg/rancher/connectionInfo.go
```diff
@@ -0,0 +1,28 @@
...
+		return "", nil
+	}
```
jakefhyde, line 18:
```suggestion
	if errors.Is(err, os.ErrNotExist) {
		logrus.Errorf("Could not find update-connection-info.ps1, will not attempt to update Rancher connection information.")
		return "", nil
	} else {
			return "", fmt.Errorf("Failed to update update-connection-info.ps1: %w", err)
	}
```

suc/pkg/rancher/connectionInfo.go
```diff
@@ -0,0 +1,28 @@
...
+
+	// This command is expected to be run in a host process pods. Files packaged into
+	// containers which are run as host process pods will be accessible from the `/hpc` directory.
+	cmd := exec.Command("powershell", "-File", connInfoScriptName)
+	o, err := cmd.CombinedOutput()
+	if err != nil {
+		return string(o), fmt.Errorf("failed to update connection info (%s): %v", string(o), err)
```
jakefhyde, line 25:
```suggestion
		return string(o), fmt.Errorf("failed to update connection info (%s): %w", string(o), err)
```

suc/main.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	app.Flags = []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "debug",
+			Usage: "Turn on verbose debug logging",
+		},
+		&cli.BoolFlag{
+			Name:  "quiet",
+			Usage: "Turn on off all logging",
```
jakefhyde, line 46:
```suggestion
			Usage: "Turn off all logging",
```

jakefhyde (changes requested):

suc/pkg/service/config.go
HarrisonWAffel: `STRICT_VERIFY` is platform agnostic and does not use a `CATTLE_*` prefix. I think supporting both formats here makes more sense than changing `STRICT_VERIFY`.

suc/pkg/service/config.go
HarrisonWAffel: Good catch, will default to lower case to match install script

suc/main.go
HarrisonWAffel: Good catch! This is also a problem in the actual `rancher-wins.exe` binary 😅

suc/update-connection-info.ps1
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+}
+
+function Update-ConnectionInfo()
+{
+    $env:RKE2_DATA_DIR = "c:\var\lib\rancher"
+    $env:CATTLE_AGENT_CONFIG_DIR = "c:\etc\rancher\wins"
+
+    $env:CATTLE_AGENT_VAR_DIR = "$env:RKE2_DATA_DIR\agent"
```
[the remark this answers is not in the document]
HarrisonWAffel: I'm a bit worried that, due to `RKE2_DATA_DIR` / `CATTLE_AGENT_VAR_DIR` not being passed properly to the system-agent-installer on initial provisioning, using it properly here might result in us using the wrong location. The intent with this PR is to isolate it to the 0.4.x release line, which will be limited to Rancher 2.9 and below. Because we don't have data dir support for windows on 2.9.x (and I'm not sure we plan on fixing that, but you can correct me if I'm wrong) I don't know if we should consider the value when updating the connection info script.

suc/pkg/service/config.go
jakefhyde: IMO I'd remove the simple variants of the functions and just add `cattle_wins_` to the definitions, makes it less confusing of when to use which.

HarrisonWAffel: Agreed

 I had initially implemented this file with the goal of allowing users to configure other areas of the rancher-wins config file, at which point I think having these functions made more sense. I walked back the idea when I realized how little of it really needs to be changed in order for us to handle the agent tls mode issue, so you're right using raw strings here is better / simpler.

suc/update-connection-info.ps1
HarrisonWAffel: I've added more logging to the various loops, let me know what you think

suc/pkg/service/config.go
HarrisonWAffel: Fair, added `getConfigPath`

suc/pkg/cmd.go
HarrisonWAffel: I've dropped the use of an error slice in this file, but you're right that would have been cleaner

suc/update-connection-info.ps1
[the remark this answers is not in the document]
HarrisonWAffel: Which variable specifically?

suc/update-connection-info.ps1
```diff
@@ -0,0 +1,172 @@
...
+                $retries++
+                continue
+            }
+        }
+    }
+}
+
+function Update-ConnectionInfo()
+{
+    $env:RKE2_DATA_DIR = "c:\var\lib\rancher"
```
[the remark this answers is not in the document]
HarrisonWAffel: Updated

HarrisonWAffel:
The latest commit includes additional logic to the `update-connection-info.ps1` script which is required for rancher-wins to properly download the csi proxy binary from the local rancher server when using private or self signed certs.

suc/pkg/cmd.go
HarrisonWAffel: I've removed the use of an error slice here, it made more sense in my POC for this change but is no longer needed

cmd/server/config/config.go
```diff
@@ -21,22 +22,19 @@ func DefaultConfig() *Config {
...
-	WhiteList   WhiteListConfig     `yaml:"white_list" json:"white_list"`
-	Upgrade     UpgradeConfig       `yaml:"upgrade" json:"upgrade"`
-	SystemAgent *config.AgentConfig `yaml:"systemagent" json:"systemagent"`
-	CSIProxy    *csiproxy.Config    `yaml:"csi-proxy" json:"csi-proxy"`
-	TLSConfig   *wintls.Config      `yaml:"tls-config" json:"tls-config"`
+	Debug              bool                `yaml:"debug" json:"debug"`
+	Listen             string              `yaml:"listen" json:"listen"`
+	Proxy              string              `yaml:"proxy" json:"proxy"`
+	WhiteList          WhiteListConfig     `yaml:"white_list" json:"white_list"`
+	SystemAgent        *config.AgentConfig `yaml:"systemagent" json:"systemagent"`
```
jiaqiluo, line 34: to keep consistence with others:
```suggestion
	SystemAgent        *config.AgentConfig `yaml:"systemagent" json:"systemagent,omitempty"`
```

cmd/server/config/config.go
```diff
@@ -122,6 +92,19 @@ func LoadConfig(path string, v *Config) error {
 	return v.Validate()
 }
 
+func SaveConfig(path string, v *Config) error {
+	if v == nil {
+		return fmt.Errorf("config cannot be nil")
```
jiaqiluo, line 97:
```suggestion
		return fmt.Error("config cannot be nil")
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+// loadConfig utilizes the ConfigDirEnvVar environment variable and path parameter to save
+// a config file on the host. If both are provided, the path parameter will take
+// precedence. If neither are provided, then the defaultConfigFile path is used.
+func saveConfig(cfg *config.Config, path string) error {
```
jiaqiluo, line 50:
```suggestion
// saveConfig utilizes the ConfigDirEnvVar environment variable and path parameter to save
// a config file on the host. If both are provided, the path parameter will take
// precedence. If neither are provided, then the defaultConfigFile path is used.
func saveConfig(cfg *config.Config, path string) error {
```

suc/pkg/service/config.go
```diff
@@ -0,0 +1,102 @@
...
+		if cfg.Debug != givenBool {
+			cfg.Debug = givenBool
+			configNeedsUpdate = true
+		}
+	}
+
+	logrus.Infof("Checking the %s value. This is a boolean flag, expecting 'true' or 'false'", AgentStringTLSEnvVar)
+	if v := os.Getenv(AgentStringTLSEnvVar); v != "" {
+		logrus.Infof("Found value '%s' for %s", v, AgentStringTLSEnvVar)
+		tlsMode, err := strconv.ParseBool(strings.ToLower(v))
```
jiaqiluo, line 81: If the expected value is "true" or "false", it is better to use the same string-compare method as the above section to avoid falsely parsing other invalid values.

suc/pkg/cmd.go
```diff
@@ -0,0 +1,51 @@
...
+		return fmt.Errorf("failed to update rancher-wins config file: %w", updateErr)
+	}
+
+	if restartServiceDueToConfigChange {
+		if err = service.RefreshWinsService(); err != nil {
+			return fmt.Errorf("error encountered while attempting to restart rancher-wins: %w", err)
+		}
+	}
+
+	return fmt.Errorf("%s encountered an error while updating the node: %w", defaults.WindowsSUCName, err)
```
jiaqiluo, line 50: Wait, does the function always return the error? 😄

suc/pkg/service/service.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to refresh the %s service: %w", winSrv.Name, err)
+		}
+	case svc.Stopped:
+		err = winSrv.Start()
+		if err != nil {
+			return fmt.Errorf("failed to start %s: %w", winSrv.Name, err)
+		}
+	default:
+		logrus.Warnf("Unexpected service state %s (%d), will not start or restart service: ", serviceStateToString(state), state)
```
jiaqiluo, line 40:
```suggestion
		logrus.Warnf("Unexpected service state %s (%d), will not start or restart service", serviceStateToString(state), state)
```

suc/pkg/service/service.go
```diff
@@ -0,0 +1,137 @@
...
+	currentState, err := getServiceState(s)
+	if err != nil {
+		return fmt.Errorf("failed to get state of service '%s': %w", s.Name, err)
+	}
+
+	if currentState != svc.Running {
+		// can't restart a service that isn't running
+		logrus.Warnf("Cannot restart the %s service as it is not yet running", s.Name)
+		return fmt.Errorf("cannot restart the %s service as it is not yet running", s.Name)
+	}
```
jiaqiluo, line 65: As the goal of the function is to ensure the service is restarted, is it better to continue to start the service, instead of erroring out, in this case?

jiaqiluo (changes requested):

cmd/server/config/config.go
HarrisonWAffel: There is no `fmt.Error`, only `fmt.Errorf`

suc/pkg/cmd.go
HarrisonWAffel: Whoops! Thought I had a return statement after refreshing the service

suc/pkg/service/config_test.go
HarrisonWAffel: These tests cannot be run in parallel as they depend on certain environment variables being present at specific times.

cmd/server/config/config.go
jakefhyde: `errors.New` perhaps?

HarrisonWAffel: Good point!

suc/pkg/service/service.go
HarrisonWAffel: My main goal was to mimic the `Restart-Service` powershell commandlet, which behaves in a similar way. This was also a requirement for sending the restart signal to the SCM. However, iirc in testing I saw that sending the restart signal via the windows package did not function in the same way as it does on Powershell, so I moved to explicit stop and start calls. 

For the purposes of handling the agent tls mode feature I agree, we can remove this conditional. Though there could be value in detecting when the service is not in the state we expect ('running') I don't think we will hit a case like that in the near future

suc/pkg/service/service.go
```diff
@@ -0,0 +1,117 @@
...
+func restartService(s *mgr.Service) error {
+	currentState, err := getServiceState(s)
+	if err != nil {
+		return fmt.Errorf("failed to get state of service %s: %w", s.Name, err)
+	}
+
+	if currentState == svc.Running {
+		if err = stopService(s); err != nil {
+			logrus.Errorf("Encountered error attempting to stop the %s service: %v", s.Name, err)
+			return fmt.Errorf("encountered error attempting to stop the %s service: %v", s.Name, err)
```
jiaqiluo, line 48:
```suggestion
			return fmt.Errorf("encountered error attempting to stop the %s service: %w", s.Name, err)
```

jiaqiluo (changes requested):

jiaqiluo (approved):
Good job!  LGTM.

suc/main.go
```diff
@@ -0,0 +1,53 @@
...
+	app.ErrWriter = colorable.NewColorableStderr()
+	app.Before = func(c *cli.Context) error {
+		if c.Bool("debug") {
+			logrus.SetLevel(logrus.DebugLevel)
+		}
+		if c.Bool("quiet") {
+			logrus.SetOutput(io.Discard)
+		}
+		logrus.SetFormatter(&logrus.TextFormatter{ForceColors: true, FullTimestamp: true})
+		logrus.SetOutput(c.App.Writer)
```
jakefhyde, line 35: Won't this overwrite the `quiet` check?

suc/pkg/cmd.go
```diff
@@ -0,0 +1,55 @@
...
+		return fmt.Errorf("failed to update rancher-wins config file: %w", updateErr)
+	}
+
+	if restartServiceDueToConfigChange {
+		if err = service.RefreshWinsService(); err != nil {
+			return fmt.Errorf("error encountered while attempting to restart rancher-wins: %w", err)
+		}
+	}
+
+	if err != nil {
```
jakefhyde, line 50: I don't think it's possible to reach this branch, every non-shadowed `err` is checked.

suc/pkg/service/config.go
```diff
@@ -0,0 +1,99 @@
...
+func getConfigPath(path string) string {
+	configDirEnv := os.Getenv(ConfigDirEnvVar)
+	if path == "" && configDirEnv != "" {
+		path = configDirEnv
+	}
+	if path == "" {
+		path = defaultConfigFile
+	}
+	return path
+}
```
jakefhyde, line 31:
```suggestion
func getConfigPath(path string) string {
  if path != "" {
    return path
  }
	if path = os.Getenv(ConfigDirEnvVar); path != "" {
	  return path
	}
	return defaultConfigFile
}
```
The first branch is somewhat confusing, but I've deduced the following:
1. If the path parameter is not empty, return that,
2. otherwise, return the value of the env var
3. otherwise, return the default

jakefhyde: It also looks like `UpdateConfigFromEnvVars` is only ever called with `""`, so do we really need the `path` parameter?

suc/pkg/service/config.go
```diff
@@ -0,0 +1,99 @@
...
+			configNeedsUpdate = true
+		}
+	}
+
+	// If we haven't made any changes there is no reason to update the config file
+	if configNeedsUpdate {
+		logrus.Info("Detected a change in configuration, updating config file")
+		err = saveConfig(cfg, path)
+		if err != nil {
+			return configNeedsUpdate, fmt.Errorf("failed to save config: %v", err)
```
jakefhyde, line 92:
```suggestion
			return configNeedsUpdate, fmt.Errorf("failed to save config: %w", err)
```

suc/pkg/service/service.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("failed to refresh the %s service: %w", winSrv.Name, err)
+	}
+
+	return nil
+}
+
+func getServiceState(s *mgr.Service) (svc.State, error) {
+	q, err := s.Query()
+	if err != nil {
+		return 0, fmt.Errorf("could not query service %s: %v", s.Name, err)
```
jakefhyde, line 33:
```suggestion
		return 0, fmt.Errorf("could not query service %s: %w", s.Name, err)
```

suc/pkg/service/service.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return waitForServiceState(s, svc.Stopped, 5, 5)
+}
+
+func waitForServiceState(s *mgr.Service, desiredState svc.State, delayInSeconds time.Duration, maxAttempts int) error {
+	transitionedSuccessfully := false
+	var err error
+	var state svc.State
+
+	for i := 0; i < maxAttempts; i++ {
```
jakefhyde, line 82:
```suggestion
	for i := range maxAttempts {
```
Not sure if you want to but you should be able to do this now.

suc/update-connection-info.ps1
```diff
@@ -0,0 +1,220 @@
...
+function Update-ConnectionInfo()
+{
+    $env:RKE2_DATA_ROOT = "c:\var\lib\rancher"
+    $env:CATTLE_AGENT_CONFIG_DIR = "c:\etc\rancher\wins"
+
+    if ($env:CATTLE_WINS_CONFIG_DIR -and ($env:CATTLE_WINS_CONFIG_DIR -ne "")) {
+        $env:CATTLE_AGENT_CONFIG_DIR = $env:CATTLE_WINS_CONFIG_DIR
+    }
+
+    $env:CATTLE_AGENT_VAR_DIR = "$env:RKE2_DATA_ROOT\agent"
```
jakefhyde, line 197:
```suggestion
    if ($env:CATTLE_AGENT_VAR_DIR -eq "") { 
      $env:CATTLE_AGENT_VAR_DIR = "$env:RKE2_DATA_ROOT\agent"
    }
```
I'd still probably do this here but it's up to you.

jakefhyde (changes requested):
Just a few nits

suc/update-connection-info.ps1
HarrisonWAffel: On 2.9 there isn't a way for a user to configure this path during the provisioning process, so exposing it and allowing a user to change it via agent environment variables won't be helpful

suc/pkg/service/config.go
HarrisonWAffel: Dropped the `path` parameter

jakefhyde (approved):

PR #247, opened 2024-09-19
[v0.4.19] Dependency bumps, fix environment variables used by mage

HarrisonWAffel:
+ Bumps several dependencies to the latest versions 
+ Fixes the PR workflow to only run on PRs 
+ Fixes the mage file to ensure the arch specified in `NewGo` is actually used. Currently, the host architecture is always being used as `ARCH` was being set instead of `GOARCH`

Notes:
+ `github.com/Microsoft/hcsshim` cannot be bumped at this time as the indirect `github.com/prometheus` dependencies rely on a type within the `k8s.io/component-base` dependency which was made private in `1.29`. This is fixed in version `v0.31.1`. When we bump all k8s dependencies to `v0.31.1` we can also bump `github.com/Microsoft/hcsshim`
+ A few other dependencies in the `replace` block have more recent versions, but have not been bumped so as to align with the latest versions used by the system-agent


jiaqiluo (approved):

HarrisonWAffel:
Merging with a single review as these are simple `go.mod` bumps and trivial updates to the mage tools

PR #248, opened 2024-09-27
[2.10] Ability to configure service start types and dependencies

HarrisonWAffel:
### Summary
Issue https://github.com/rancher/rancher/issues/42458

This PR refactors the SUC image with the aim of providing new ways to configure the `rancher-wins` and `rke2` services. Namely, these changes introduce the ability to control the start type of the `rancher-wins` Windows service and to define a service dependency between `rancher-wins` and `rke2`. 

These features can be controlled via environment variables, typically set by Rancher when deploying SUC plans. These variables can be provided in the Rancher UI by setting the following environment variables in the `Agent Environment Variables` section. 

To change the `rancher-wins` start type to delayed, the `CATTLE_ENABLE_WINS_DELAYED_START` can be set to `true`. To add `rancher-wins` as a service dependency on `rke2`, the `CATTLE_ENABLE_WINS_SERVICE_DEPENDENCY` can be set to `true`. If a variable is omitted, it is treated as if it were disabled, reverting any changes made previously. If any errors are encountered during this process, any changes made will be reverted back to their initial state.

New integration tests have been created to cover this functionality, the Go version has been bumped, and the CI updated. 

Provisioning and configuration via Rancher.


suc/pkg/service/configure.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+		if hasDep {
+			logrus.Info("Temporarily removing rke2 service dependency")
+			depRemoved = true
+			err = rke2Srv.RemoveRancherWinsServiceDependency()
+			if err != nil {
+				return fmt.Errorf("error encountered while temporarily removing rke2 service dependency: %w", err)
+			}
+		}
+		rke2Srv.Close()
```
jiaqiluo, line 130: It is better to close the service before checking the error, otherwise, the service will not be closed if the error is not nil. 
```suggestion
			err = rke2Srv.RemoveRancherWinsServiceDependency()
			rke2Srv.Close()
			if err != nil {
				return fmt.Errorf("error encountered while temporarily removing rke2 service dependency: %w", err)
			}
		}

```

suc/pkg/service/configure.go
```diff
@@ -0,0 +1,153 @@
...
+		rke2Srv, _, err = OpenRke2Service()
+		if err != nil {
+			logrus.Errorf("error opening rke2 service while restarting rancher-wins: %v", err)
+		}
+		logrus.Info("Restoring rke2 service dependency")
+		err = rke2Srv.AddRancherWinsServiceDependency()
+		if err != nil {
+			return fmt.Errorf("error encountered while restoring rke2 service dependency: %w", err)
+		}
+		rke2Srv.Close()
```
jiaqiluo, line 149: For the same reason as the above comment 
```suggestion
		err = rke2Srv.AddRancherWinsServiceDependency()
		rke2Srv.Close()
		if err != nil {
			return fmt.Errorf("error encountered while restoring rke2 service dependency: %w", err)
		}
```

suc/pkg/service/service.go
```diff
@@ -1,86 +1,132 @@
...
 	}
 
-	if err = s.Start(); err != nil {
-		return fmt.Errorf("failed to restart the %s service: %w", s.Name, err)
+	if err = s.svc.Start(); err != nil {
+		return fmt.Errorf("failed to start the %s service while attempting to restart: %w", s.Name, err)
 	}
 
-	return waitForServiceState(s, svc.Running, 5, 5)
+	return s.WaitForState(svc.Running, 5, 5)
```
jiaqiluo, line 100: As the same values for `delayInSeconds` and `maxAttempts` are used at multiple places, it is better to define the values as package-level constants.

jiaqiluo (changes requested):
I left some comments, but overall it looked good to me.

suc/pkg/service/common.go
```diff
@@ -0,0 +1,48 @@
+package service
+
+import (
+	"golang.org/x/sys/windows/svc"
+)
+
+func SlicesMatch(s1 []string, s2 []string) bool {
```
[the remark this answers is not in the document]
HarrisonWAffel: Good catch!

cmd/server/app/run.go
```diff
@@ -41,6 +41,11 @@ var _runFlags = []cli.Flag{
 		Usage: "[optional] Specifies the name of the file to write the profile to",
 		Value: "profile.pprof",
 	},
+	&cli.BoolFlag{
+		Name:  "delayed-start",
+		Usage: "configure the rancher-wins service with a start type of 'Automatic (Delayed)'",
```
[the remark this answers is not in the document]
jiaqiluo: It looks like the convention is to add "[optional]" to the Usage line :)

jiaqiluo (changes requested):

cmd/server/app/run.go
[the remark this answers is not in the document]
HarrisonWAffel: Need more coffee 😅

jiaqiluo (approved):
Good Job! 
LGTM 💯

suc/pkg/service/common.go
```diff
@@ -0,0 +1,51 @@
...
```
jakefhyde, line 7: This function is only used once, is it written this way because the ordering doesn't matter? I feel like if the ordering is always the same and does matter, it would be easier to use `reflect.DeepEqual` or compare each index of the slice. 

In other words, if the time complexity of this function is on purpose, it should have a comment and test cases and maybe update the name to `UnorderedSlicesEqual` or something similar (honestly I would use generics for this but that's just me), and it it's not, you probably don't need the function at all and could use `reflect.DeepEqual` wherever this function is called.

suc/pkg/service/service_rke2.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/rancher/wins/pkg/defaults"
+	"github.com/sirupsen/logrus"
+)
+
+type Rke2Service struct {
+	Service
+}
+
+func OpenRke2Service() (*Rke2Service, bool, error) {
```
jakefhyde, line 14:
```suggestion
func OpenRKE2Service() (*RKE2Service, bool, error) {
```
https://go.dev/wiki/CodeReviewComments#initialisms

suc/pkg/service/service_rke2.go
```diff
@@ -0,0 +1,56 @@
+package service
+
+import (
+	"fmt"
+
+	"github.com/rancher/wins/pkg/defaults"
+	"github.com/sirupsen/logrus"
+)
+
+type Rke2Service struct {
```
jakefhyde, line 10:
```suggestion
type RKE2Service struct {
```

suc/pkg/service/service_rancher_wins.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/rancher/wins/pkg/defaults"
+	"github.com/sirupsen/logrus"
+)
+
+type RancherWinsService struct {
+	Service
+}
+
+func OpenRancherWinsService() (*RancherWinsService, bool, error) {
+
```
jakefhyde, line 15:
```suggestion
```

suc/pkg/service/state/state.go
```diff
@@ -0,0 +1,126 @@
...
+	if rke2Exists {
+		rke2Deps = rke2Svc.Config.Dependencies
+		rke2Svc.Close()
+	} else {
+		logrus.Warn("Could not find rke2 service while building initial state")
+	}
+
+	logrus.Debugf("rancher-wins delayed start is set to %t", winsSvc.Config.DelayedAutoStart)
+	logrus.Debugf("rke2 service dependencies: %v", rke2Deps)
+	j, _ := json.MarshalIndent(winsCfg, "", " ")
```
jakefhyde, line 67: I'd still probably return this error.

tests/integration/utils.psm1
```diff
@@ -182,6 +179,71 @@ function New-Directory {
     }
 }
 
+function Add-DummyRke2Service {
```
jakefhyde, line 182:
```suggestion
function Add-DummyRKE2Service {
```

tests/integration/utils.psm1
```diff
@@ -182,6 +179,71 @@ function New-Directory {
...
+  }
+"@
+    # Compile the above c# into a binary that can be used as a service.
+    # Note: This service will not start, as there is no service handler implemented, however it can still
+    # be configured in the same ways that any other functional service can be.
+    Add-Type -TypeDefinition $dummyServiceScript -Language CSharp -OutputAssembly "rke2.exe" -OutputType ConsoleApplication
+    New-Service -Name "rke2" -BinaryPathName "rke2.exe"
+}
+
+function Remove-DummyRke2Service {
```
jakefhyde, line 199:
```suggestion
function Remove-DummyRKE2Service {
```

suc/pkg/service/state/state.go
HarrisonWAffel: Updated

suc/pkg/service/common.go
HarrisonWAffel: I couldn't find documentation that guaranteed the windows api would return dependencies in the same order each time, so I'd rather leave this as unordered. I've updated this to use generics and I've added a unit test case

suc/pkg/service/common.go
```diff
@@ -0,0 +1,48 @@
...
+
+func removeFromSlice(x string, s []string) []string {
+	var n []string
+	for _, e := range s {
+		if e == x {
+			continue
+		}
+		n = append(n, e)
+	}
+	return n
```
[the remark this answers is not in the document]
HarrisonWAffel: Renamed to `removeAllFromSlice` and refactored to use generics

suc/pkg/rancher/connectionInfo.go
```diff
@@ -11,21 +11,21 @@ import (
 
 const connInfoScriptName = "/hpc/update-connection-info.ps1"
 
-func UpdateConnectionInformation() (string, error) {
+func UpdateConnectionInformation() (string, bool, error) {
```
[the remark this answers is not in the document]
HarrisonWAffel: Updated

suc/pkg/rancher/connectionInfo.go
```diff
@@ -27,5 +27,7 @@ func UpdateConnectionInformation() (string, error) {
 	if err != nil {
 		return string(o), fmt.Errorf("failed to update connection info: %w", err)
 	}
+
+	logrus.Info("Successfully updated connection info")
```
jakefhyde, line 31: When should a log message end with a period and when shouldn't it? I won't block on this, just food for thought.

suc/pkg/service/service_rke2.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/rancher/wins/pkg/defaults"
+	"github.com/sirupsen/logrus"
+)
+
+type RKE2Service struct {
+	Service
+}
+
+func OpenRke2Service() (*RKE2Service, bool, error) {
```
jakefhyde, line 14:
```suggestion
func OpenRKE2Service() (*RKE2Service, bool, error) {
```

jakefhyde (changes requested):
One formatting nit

suc/pkg/service/service_rke2.go
HarrisonWAffel: Good catch

jakefhyde (approved):

HarrisonWAffel:
Merging this as the only additional commit added past initial review is bumping of the `system-agent` dependency.

PR #260, opened 2024-11-14
Implement auto upgrades of wins.exe via the SUC image

HarrisonWAffel:
### Summary
Issue: https://github.com/rancher/rancher/issues/47652

### Problem
The `rancher-wins` system-upgrade-controller image does not automatically upgrade the `wins.exe` binary on the host. This forces users to reprovision Windows nodes in order to obtain the latest version of the `system-agent` and `rancher-wins`, and has resulted in unnecessary changes in Rancher to handle behavior differences between different versions of `rancher-wins`. 

### Solution

This PR introduces a new `host` package which embeds the latest version of `wins.exe` into the SUC image during compilation. When the SUC image runs, it will compare the version of wins installed onto the host (by invoking `c:\Windows\wins.exe --version`) and, if needed, write the embedded version to disk into the required directories (`c:\Windows` & `c:\usr\local\bin`). 

This behavior can be disabled by setting the `CATTLE_WINS_SKIP_BINARY_UPGRADE` environment variable to `true`. This is required when running CI tests which do not fully install `rancher-wins`. 

CI has been updated to cover this behavior during PRs and releases, ensuring that the proper version of `wins.exe` is always embedded into the SUC binary. Additionally, existing tests have been adjusted to ensure that we are tracking the proper exit code of certain commands. 

### Extra Context 

This PR does _not_ handle automatic upgrades of the csiproxy. Additional investigation needs to be done regarding this component.

### DIFF
--- a/suc/pkg/host/common.go
+++ b/suc/pkg/host/common.go
@@ -1,0 +1,151 @@
+package host
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	defaultWinsPath                    = "c:\\Windows\\wins.exe"
+	defaultWinsUsrLocalBinPath         = "c:\\usr\\local\\bin\\wins.exe"
+	defaultConfigDir                   = "c:\\etc\\rancher\\wins"
+	fileOperationAttempts              = 5
+	fileOperationAttemptDelayInSeconds = 5
+
+	// skipBinaryUpgradeEnvVar prevents the suc image from attempting to upgrade the wins binary.
+	// This is primarily used in CI, to allow for test cases to run without having to completely
+	// install rancher-wins.
+	skipBinaryUpgradeEnvVar = "CATTLE_WINS_SKIP_BINARY_UPGRADE"
+)
+
+// getRancherWinsVersionFromBinary executes the wins.exe binary located at 'path' and passes the '--version'
+// flag. The release version or commit hash is returned. If the binary returns unexpected output,
+// was built with a dirty commit, or does not exist, an error will be returned.
+func getRancherWinsVersionFromBinary(path string) (string, error) {
+	if path == "" {
+		return "", fmt.Errorf("must specify a path")
+	}
+
+	_, err := os.Stat(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return "", fmt.Errorf("provided path (%s) does not exist", path)
+		}
+		return "", fmt.Errorf("encoutered error stat'ing '%s': %w", path, err)
+	}
+
+	out, err := exec.Command(path, "--version").CombinedOutput()
+	if err != nil {
+		logrus.Errorf("could not invoke '%s --version' to determine installed wins.exe version: %v", path, err)
+		return "", fmt.Errorf("failed to invoke '%s --version': %w", path, err)
+	}
+
+	logrus.Debugf("'%s --version' output: %s", path, string(out))
+	return parseWinsVersion(string(out))
+}
+
+func confirmWinsBinaryVersion(desiredVersion string, path string) error {
+	installedVersion, err := getRancherWinsVersionFromBinary(path)
+	if err != nil {
+		return fmt.Errorf("failed to confirm '%s' version: %w", path, err)
+	}
+
+	if installedVersion == desiredVersion {
+		logrus.Debugf("'%s' returned expected version (%s)", path, desiredVersion)
+		return nil
+	}
+
+	return fmt.Errorf("'%s' version ('%s') did not match desired version ('%s')", path, installedVersion, desiredVersion)
+}
+
+func parseWinsVersion(winsOutput string) (string, error) {
+	// Expected output format is 'rancher-wins version v0.x.y[-rc.z]'"
+	// A dirty binary will return 'rancher-wins version COMMIT-dirty'
+	// A non-tagged version will return 'rancher-wins version COMMIT'
+	s := strings.Split(winsOutput, " ")
+	if len(s) != 3 {
+		return "", fmt.Errorf("'wins.exe --version' did not return expected output length ('%v' was returned)", s)
+	}
+
+	verString := strings.Trim(s[2], "\n")
+	// We should error out if the binary we're working with is dirty, but
+	// if it's simply untagged we should proceed with the upgrade.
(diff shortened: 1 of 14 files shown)


suc/pkg/host/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+	err = confirmWinsBinaryVersion(desiredVersion, updatedBinaryPath)
+	if err != nil {
+		return false, fmt.Errorf("failed to stage updated binary: %w", err)
+	}
+
+	logrus.Info("Stopping rancher-wins...")
+	rw, _, err := service.OpenRancherWinsService()
+	if err != nil {
+		return false, fmt.Errorf("failed to open rancher-wins service while attempting to upgrade binary: %w", err)
+	}
```
jiaqiluo, line 78: This might be a silly question, but is it possible that an error returned because the service doesn't exist, especially in the case of a fresh installation? I'm asking to ensure that the logic doesn't introduce any unnecessary failures.

suc/pkg/host/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+		return false, fmt.Errorf("failed to copy new wins.exe binary to %s: %w", usrLocalBinPath, err)
+	}
+
+	logrus.Infof("Validating updated binaries...")
+	err = confirmWinsBinaryVersion(desiredVersion, defaultWinsPath)
+	if err != nil {
+		return false, err
+	}
+
+	err = confirmWinsBinaryVersion(desiredVersion, getWinsUsrLocalBinBinary())
```
jiaqiluo, line 110: nit
```suggestion
	err = confirmWinsBinaryVersion(desiredVersion,usrLocalBinPath)
```

suc/pkg/host/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	logrus.Infof("Removing %s", updatedBinaryPath)
+	err = os.Remove(updatedBinaryPath)
```
jiaqiluo, line 116: Why are we removing the source binary here? Could it be possible that the binary at the target location gets removed somehow, and we would need to copy it over again?

jiaqiluo (changes requested):
overall LGTM, left a few comments to confirm some information.

suc/pkg/host/common.go
```diff
@@ -0,0 +1,151 @@
...
+	// install rancher-wins.
+	skipBinaryUpgradeEnvVar = "CATTLE_WINS_SKIP_BINARY_UPGRADE"
+)
+
+// getRancherWinsVersionFromBinary executes the wins.exe binary located at 'path' and passes the '--version'
+// flag. The release version or commit hash is returned. If the binary returns unexpected output,
+// was built with a dirty commit, or does not exist, an error will be returned.
+func getRancherWinsVersionFromBinary(path string) (string, error) {
+	if path == "" {
+		return "", fmt.Errorf("must specify a path")
```
jakefhyde, line 32: IMO I prefer declaring errors as private package variables with `errors.New` so it's easy to write unit tests for them with `errors.Is`. You have some tests for this function that just expect an error being returned, and I know that realistically there isn't much of a use to do so, but personally I like to check the exact error that is being returned if at all possible. I won't block on this, but I've found it's generally a good practice, although it can conflict with writing table style tests.

suc/pkg/host/common.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	_, err := os.Stat(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return "", fmt.Errorf("provided path (%s) does not exist", path)
```
jakefhyde, line 38: Just out of curiosity, what is the purpose for returning this separately as opposed to just using the following return?

suc/pkg/host/upgrade.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	logrus.Info("Stopping rancher-wins...")
+	rw, _, err := service.OpenRancherWinsService()
+	if err != nil {
+		return false, fmt.Errorf("failed to open rancher-wins service while attempting to upgrade binary: %w", err)
+	}
+
+	// The service needs to be stopped before we can modify
+	// the binary it uses
```
jakefhyde, line 81:
```suggestion
	// The service needs to be stopped before we can modify the binary it uses
```

suc/pkg/host/upgrade.go
HarrisonWAffel: That's a good point but my understanding is that scenario can't happen. The node needs to be connected to the cluster before any SUC plans are delivered, which requires the rancher-wins service to be running. However I can add some additional handling just to be safe. In the scenario you're describing the `OpenRancherWinsService` wouldn't fail, but `rw.Stop` would.

jakefhyde (commented):
Just a few minor nits, nothing blocking.

suc/pkg/host/upgrade.go
HarrisonWAffel: The intent here was to ensure that we aren't leaving binaries on the host that aren't being used. Do you mean that it would be useful for users to have access to this file in case they need to manually restore to that version? 

Technically there _are_ two copies of wins.exe on each node, so there is still a backup that can be used to manually update the node if one is deleted. lmk if I've misunderstood your comment though 😄

suc/pkg/host/upgrade.go
jiaqiluo: sgtm

suc/pkg/host/upgrade.go
jiaqiluo: I was initially considering the case where the function might be invoked again and fail to find the binary because it was removed. However, I now realize that the binary is recreated each time the function is called, so my question is no longer relevant.

jiaqiluo (approved):

suc/pkg/host/common.go
HarrisonWAffel: It's been a while since i first raised this PR, but iirc this resulted in a more useful/readable error message when inspecting the SUC pod logs

suc/pkg/host/common.go
HarrisonWAffel: I agree with the sentiment, however this particular function is tested via the pester integration tests, so we wouldn't be able to leverage new error specific variables. Additionally a lot of the errors include runtime specific information (like paths and whatnot) that we don't know at compile time.

jakefhyde (approved):

PR #301, opened 2025-07-22
chore: un-rc dependencies for 2.12 release

HarrisonWAffel:
un-rc'ing dependencies for 2.12 release 

This also addresses https://github.com/rancher/wins/pull/300


go.mod
```diff
@@ -60,8 +59,8 @@ require (
 	github.com/magefile/mage v1.15.0
 	github.com/mattn/go-colorable v0.1.13
 	github.com/pkg/errors v0.9.1
-	github.com/rancher/remotedialer v0.4.5-rc.2
-	github.com/rancher/system-agent v0.3.13-rc.3
+	github.com/rancher/remotedialer v0.5.0
```
HarrisonWAffel, line 62: Just want to call out that the `remotedialer` dependency is not used in the current provisioned flow. This is only used when defining proxies via wins, which is an RKE1 specific behavior that is no longer used in any capacity.

jakefhyde (approved):

go.mod
```diff
@@ -12,7 +12,6 @@ replace (
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.58.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.58.0
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc => go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.33.0
-	golang.org/x/oauth2 => golang.org/x/oauth2 v0.25.0
```
jiaqiluo, line 15: Is it no longer needed to ping it to a specific tag?  I noticed the auto-generated PR bumps the tag to  v0.27.0

jiaqiluo (changes requested):

go.mod
HarrisonWAffel: The [indirect dependency](https://github.com/rancher/wins/blob/main/go.mod#L160C2-L160C29) was already pointing to `v0.27.0`, and ideally this would always match what is being used by the system-agent.

But I don't mind keeping the explicit version pinning, I've added the replace statement back.

jiaqiluo (approved):

pedromfcarvalho (approved):

PR #312, opened 2025-10-07
Bump system-agent to v0.3.14-rc.3 and K8s to v1.33.5

macedogm:
Bump system-agent to `v0.3.14-rc.3` and K8s `to v1.33.5` to remove 2 CVEs.


HarrisonWAffel:
Approving with a single review as these are simple dependency bumps

PR #336, opened 2026-03-16
Uninstall script improvements

vardhaman22:
### Summary

Issue: https://github.com/rancher/rancher/issues/51850
addresses the following points from the issue
- The parameters listed in the script are not needed, and can cause confusion when looking at the file for the first time
- The script does not explicitly stop the csi-proxy process, and may fail to remove the related service on first run
- It may be beneficial to automatically place this script onto nodes provisioned via Rancher, so that users can easily leverage it
- Safe guards could be implemented to prevent the uninstall script from executing if rke2 is running, as wins is primarily used in the Rancher provisioning process

**updates in the install.ps1 script**
1. it will now download and mount the uninstall.ps1 script on the node with the name `wins-agent-uninstall.ps1`. there is an environment variable to specify local uninstall.ps1 as well and `CATTLE_AGENT_UNINSTALL_URL` can be used to set the download url else `$env:CATTLE_AGENT_BINARY_BASE_URL/wins-agent-uninstall.ps1` is used. If no environment is provided then it will download from github release of the specified wins version
2. removed the logic of downloading the latest available wins.exe on github and added checksum verification for wins and uninstall script both.

**updates in uninstall.ps1 script**
1.  it will fail if rke2 process or service is running on the node.
2. removed unnecessary params from the script.
3. it will stop cisproxy service before wins service
4. it will stop csi-proxy and wins processes after stopping the services
5. check for service existence before deleting the service

**uninstall.ps1 tests"
1. added integration tests for uninstall.ps1 script.
for the tests created a test service in test/integration/testservice which will be used to generate an executable (test.exe) and that executable will be used to create dummy processes and services(rancher-wins and csi-proxy). Those will be then cleaned up by uninstall.ps1 script while testing.

### Technical notes summary

the install.sh script will now auto mount the `uninstall.ps1` script on the nodes with the name `wins-agent-uninstall.ps1`.
and uninstall.ps1 script will now remove `cisproxy` service and the processes associated with `wins` and `cisproxy` services. 

- provisioning rke2 clusters with windows worker nodes. since install.sh script is changed.

- provisioning rke2 clusters with windows worker nodes

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -20,13 +20,14 @@
       - CATTLE_LABELS
       - CATTLE_TAINTS
       Advanced Environment Variables
-      - CATTLE_AGENT_BINARY_URL (default: latest GitHub release)
+      - CATTLE_AGENT_BINARY_URL (default: pinned GitHub release)
       - CATTLE_PRESERVE_WORKDIR (default: false)
       - CATTLE_REMOTE_ENABLED (default: true)
       - CATTLE_LOCAL_ENABLED (default: false)
       - CATTLE_ID (default: autogenerate)
       - CATTLE_AGENT_BINARY_LOCAL (default: false)
       - CATTLE_AGENT_BINARY_LOCAL_LOCATION (default: )
+      - CATTLE_AGENT_BINARY_CHECKSUM (default: hardcoded fallback checksum)
       - CSI_PROXY_URL (default: )
       - CSI_PROXY_VERSION (default: )
       - CSI_PROXY_KUBELET_PATH (default: )
@@ -73,6 +74,7 @@ $ErrorActionPreference = 'Stop'
 Set-StrictMode -Version Latest
 
 $FALLBACK = "v0.4.15"
+$FALLBACK_BINARY_SUM = "8bb68f9e3d92ac08be66cb5fbe6d1dce5c4460b26e0b033e58fece21ce9a8680"
 
 function Invoke-WinsInstaller {
     [CmdletBinding()]
@@ -139,6 +141,19 @@ function Invoke-WinsInstaller {
         $writer.Flush()
         $stringAsStream.Position = 0
         return (Get-FileHash -InputStream $stringAsStream -Algorithm SHA256).Hash.ToLower()
+    }
+
+    function Set-FallbackVersion {
+        param (
+            [Parameter(Mandatory=$true)]
+            [string]
+            $Reason
+        )
+        Write-LogInfo $Reason
+        $env:VERSION = $FALLBACK
+        if (-Not $env:CATTLE_AGENT_BINARY_CHECKSUM) {
+            $env:CATTLE_AGENT_BINARY_CHECKSUM = $FALLBACK_BINARY_SUM
+        }
     }
 
     function Get-Args {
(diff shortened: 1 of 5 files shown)


HarrisonWAffel (changes requested):
A few nits and minor changes. Also, could you look into possibly expanding the integration tests to test the uninstall script explicitly? I know we are already using it some tests to clean the GH runner, but there may be a few other cases now we can test

install.ps1
```diff
@@ -289,6 +292,45 @@ function Invoke-WinsInstaller {
...
+            if (-Not $env:CATTLE_AGENT_UNINSTALL_URL) {
+                if ($env:VERSION) {
+                    Write-LogInfo "Version $env:VERSION used for downloading the wins binary, will reuse for uninstall script"
+                }
+                else {
+                    $rateLimit = $(curl.exe --connect-timeout 60 --max-time 300 $env:CURL_CAFLAG -sfL "https://api.github.com/rate_limit") | ConvertFrom-Json
+
+                    if ($rateLimit.rate.remaining -eq 0) {
+                        Write-LogInfo "Error contacting GitHub to retrieve the latest version, falling back to version: $FALLBACK"
+                        $env:VERSION = $FALLBACK
```
HarrisonWAffel: With the recent hardening going on we'll also need to ensure that the fallback version we downloaded matches the expected SHASUM, similar to how the linux system agent is handling this logic https://github.com/rancher/system-agent/pull/271

HarrisonWAffel (commented):
New integration tests look great! Only one extra change is needed due to the recent hardening across repos, specifically with how we are pulling the fallback version

install.ps1
vardhaman22: done i have added checksum verification and removed logic of downloading the latest available verison

install.ps1
```diff
@@ -268,27 +291,34 @@ function Invoke-WinsInstaller {
             }
 
             if (-Not $env:CATTLE_AGENT_BINARY_URL) {
-                $rateLimit = $(curl.exe --connect-timeout 60 --max-time 300 $env:CURL_CAFLAG -sfL "https://api.github.com/rate_limit") | ConvertFrom-Json
```
HarrisonWAffel, line 271: Why get rid of the rate limiting check? I get that it's not 100% necessary, but it does result in a cleaner error message for users as opposed to the later curl request simply failing.

install.ps1
```diff
@@ -342,48 +372,98 @@ function Invoke-WinsInstaller {
...
+
+        if (-Not (Test-Path $DestinationPath)) {
+            Write-LogFatal "$Name doesn't appear to have been installed at $DestinationPath"
+        }
+
+        if ($Checksum) {
+            Write-LogInfo "Verifying checksum for $Name"
+            $actual = (Get-FileHash -Path $DestinationPath -Algorithm SHA256).Hash.ToLower()
+            if ($actual -ne $Checksum.ToLower()) {
+                Write-LogFatal "Checksum validation failed for $Name.`n  Expected: $Checksum`n  Got:      $actual"
```
HarrisonWAffel:
```suggestion
                Write-LogFatal "Checksum validation failed for $Name.`n  Expected: $Checksum`n  Got: $actual"
```
Did you find that  these spaces were needed to make the message easier to read when you were testing these changes?

HarrisonWAffel (commented):
two comments / nits, otherwise lgtm

install.ps1
vardhaman22: sorry about this..left it by mistake
also removed the expected and got from log message since now i think only a message saying verification failed should be enough.

install.ps1
vardhaman22: rate limiting check was only necessary for checking the latest version from git...
for the else part (i.e. rate limiting reached and we use the fallback version )it was not in use so removed it.
only the fallback version will be used from now on and we will not query github for the latest release.

HarrisonWAffel: Ah I see that now, thanks

install.ps1
HarrisonWAffel: No worries, looks good

HarrisonWAffel (approved):

thatmidwesterncoder (approved):

HarrisonWAffel:
merging into main, this won't be backported into `release/v2.15` until v2.15.0 is actually out

PR #352, opened 2026-04-15
setup release workflow for supporting immutable releases

thatmidwesterncoder:


.github/workflows/release.yaml
```diff
@@ -6,7 +6,20 @@ on:
       - '*'
 
 jobs:
+  create-draft:
```
HarrisonWAffel: Can we move this step later on in the workflow? In the event the CI fails (which is happening more often due to changes in the GHA windows runners) we would end up trying to create a draft release several times. Not sure if that would fail or succeed. 

Also, do immutable releases just require us to always create a draft release before an official release?

HarrisonWAffel (changes requested):

.github/workflows/release.yaml
thatmidwesterncoder: That's a good point. So I did a check locally on my fork of rancher and we can have multiple with the same name: 
image

so that is a good point we will end up with a "bad" release on failure...I wonder if it's worth coming up with a general solution that will clean up ci-created draft releases.

---

As far as creating the draft, uploading, then publishing it's just a best practice: https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases#best-practices-for-publishing-immutable-releases

my guess (uneducated) is that if the asset upload fails during creation it would be immutable and we wouldn't be able to fix it. the draft, upload, release flow ensures the assets are there.

HarrisonWAffel: Okay, makes sense. Yeah I don't really have a problem with making the draft itself, was mostly just curious. I think in the case of this repo, we don't need a solution for cleaning up bad drafts, maybe just deferring the creation of the draft until the integration tests would be enough

HarrisonWAffel (approved):

.github/workflows/release.yaml
```diff
@@ -163,3 +186,45 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+    steps:
+      - name: Create draft release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release create -R "${GITHUB_REPOSITORY}" --draft --generate-notes "${GITHUB_REF_NAME}"
```
jiaqiluo: There is a `create-draft` step above, why do we do it again here?

jiaqiluo (changes requested):

.github/workflows/release.yaml
HarrisonWAffel: Oh good catch, @thatmidwesterncoder is this actually required?

thatmidwesterncoder: :facepalm: 100% mistake on my part. must have copied it then pasted it twice due to getting distracted mid-update. It's fixed now.

jiaqiluo (approved):

PR #391, opened 2026-07-06
2.15 bumps

HarrisonWAffel:
This PR bumps the system-agent dependency to the latest RC, which resolves several CVEs. It also combines numerous other renovate PRs into two commits, so that they can be more easily backported. Going forward renovate will now open PRs for each both current release branches, so we can just merge those directly. 

I'll backport the relevant commits into the other release branches next.


jiaqiluo (approved):

HarrisonWAffel:
merging with a single review as these are simple bumps / package changes

PR #394, opened 2026-07-06
Improve handling of certificate chains

HarrisonWAffel:
### Summary
Issue: https://github.com/rancher/rancher/issues/55817

The current `install.ps1` uses a fairly fragile regex to parse certificate chains returned from the `/cacerts` endpoint. Under normal circumstances, where the Rancher configuration strictly follows the documentation, the current implementation works as expected. However, it does not handle slight misconfigurations, such as errant new lines, or the inclusion of other PKI information such as private keys. While both of those scenarios differ from the provided documentation and should be considered misconfigurations, `curl` is able to more gracefully handle these cases. 

This PR improves the regex used to break the obtained certificate chain into independent entries and import them individually into the windows certificate store. It also improves how the wins agent binary is executed, to return more detailed error logging if the binary fails to start up. Integration tests are also expanded to handle the new certificate logic. 

Creating custom or provisioned windows nodes in Rancher environments which use custom self signed certificated, served via `/cacerts`

Windows node registration / provisioning

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -500,42 +500,56 @@ function Invoke-WinsInstaller {
     # Import-Certificates imports one or more
     # certificates stored in the file specified by $ENV:RANCHER_CERT.
     # In the event that $env:RANCHER_CERT contains multiple certificate
-    # blocks, a regular expression is used to iterate across all blocks
-    # and import them individually.
+    # blocks, a regular expression is used to iterate across all valid x509 blocks
+    # and import them individually. Errant non-certificate blocks and any lingering
+    #  whitespace are ignored.
     function Import-Certificates() {
-        $rawCerts = Get-Content $env:RANCHER_CERT -Raw
-        # splits blocks, handles
-        # CLRF line endings if present
-        $pattern  = "(?<=-\r?\n)(?=-+)"
-        $allCerts = $rawCerts -split $pattern
-
-        if ($allCerts.Length -eq 0) {
-            Write-LogWarn "No Certs found in certs file, check that $env:RANCHER_CERT exists and that the correct certificate is configured at $( $env:CATTLE_SERVER )/cacerts."
+        if (-Not (Test-Path $env:RANCHER_CERT)) {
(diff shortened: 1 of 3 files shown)


vardhaman22 (approved):

install.ps1
```diff
@@ -897,35 +953,7 @@ csi-proxy:
...
-            Start-Sleep -s 5
-        }
-
-        try
-        {
-            Write-LogInfo "Starting $serviceName service."
-            Start-Service -Name $serviceName
-        } catch {
-            Write-LogInfo "$serviceName failed to start. Check the $serviceName logs for more information"
-            Write-LogInfo "Command: Get-WinEvent -ProviderName $serviceName | select-object TimeCreated,Message | Format-Table -wrap"
```
jiaqiluo, line 922: nit: 
Add back those two messages, or similar ones, to make the logs a bit more user-friendly.

jiaqiluo (changes requested):
LGTM overall. Please consider addressing my and Copilot's comments.

install.ps1
HarrisonWAffel: This command was actually not useful, as on an initial install the event log would always be empty.

install.ps1
```diff
@@ -672,7 +686,49 @@ csi-proxy:
...
+                if ($env:CATTLE_ENABLE_WINS_DELAYED_START -eq "true") {
+                    wins.exe srv app run --register --delayed-start
+                } else {
+                    wins.exe srv app run --register
+                }
+            } catch {
+                Write-LogFatal "Failed to execute wins.exe: $($_.Exception.Message)"
+            }
+            Pop-Location
+            Start-Sleep -s 5
```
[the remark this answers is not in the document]
HarrisonWAffel: To be clear, the goal here was more to catch when the exe itself was corrupted / entirely un-executable, not just when it returns a non-zero exit code. Any error encountered after the exe is successfully invoked will be logged [at the fatal level](https://github.com/rancher/wins/blob/main/cmd/main.go#L91), so it should already be included in the install script output and the ErrorActionPreference will ensure the script exits after that point.

jiaqiluo (approved):

PR #395, opened 2026-07-06
fix checksums in release

HarrisonWAffel:
also publish install and uninstall ps scripts in release

(cherry picked from commit d23c722c78fc70fbb45a78efac36a814c7b596a4)


jiaqiluo (approved):
changes match the source commit

PR #466, opened 2026-07-31
[2.14] Normalize line endings in install and uninstall scripts

jiaqiluo:
Backport https://github.com/rancher/wins/commit/6dfbf2111f93da4bcec7294c826495f7ee953bca, which is needed to make sure that the install and uninstall script artifacts do not use carriage return line endings. 
 
We will need to cut a new RC tag for the updatecli in r/r to pick up the correct values.

### DIFF
--- a/magefiles/magefile.go
+++ b/magefiles/magefile.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


HarrisonWAffel (approved):
LGTM, simple backport

PR #469, opened 2026-08-04
[release/v2.15] removed unnecessary integration tests

vardhaman22:
### Summary

Issue: there are some integration tests which were only relevant for rke1 and since it is EOL so those can be removed.
also ci sometimes fails because docker has not started before running the integration tests.

- this PR adds a section in the integration test suite script to wait for docker to start before executing any docker commands and also removes the unnecessary integration tests


HarrisonWAffel (approved):

HarrisonWAffel:
We should port this to main and older release lines too

